PR #4, opened 2022-05-24
Switch to SWC; don't crash if there's no data

brandones:


brandones:
Yep, thanks for the heads up. Looks like I needed to [update the framework](http://o3-dev.docs.openmrs.org/#/main/faq?id=im-not-seeing-the-latest-openmrsesm-framework-how-do-i-update-the-dependency).

PR #7, opened 2022-08-15
O3-1216 dashboard list accordion

pirupius:
https://user-images.githubusercontent.com/15266028/184661765-48c1206f-f833-4d30-847f-c42b16594f19.mov


package.json
```diff
@@ -57,7 +57,7 @@
     "rxjs": "6.x"
   },
   "devDependencies": {
-    "@openmrs/esm-framework": "next",
+    "@openmrs/esm-framework": "latest",
```
pirupius, line 60: changed to `latest` because `next` has the `4.0 (React 18, Router 6 and Carbon 11)` changes merged in.

cioan (approved):
LGTM! Thanks @pirupius !

src/components/prescription-details.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { medications, isError, isLoading } = useOrderDetails(encounterUuid);
+
+  return (
+    <div className={styles.prescriptionContainer}>
+      <Tile className={styles.allergiesTile}>
+        <div className={styles.allergesContent}>
+          <div>
+            <WarningFilled24 className={styles.allergiesIcon} />
+            <p>
+              <b>3 allergies</b> Penicillin, Naproxen sodium, Ibuprofen
```
mseaton, line 22: We need to be careful about inserting dummy data into these components, especially data like this.

mseaton (approved):
Generally looks great to me, nice work!  Only one comment about dummy data, might want to replace that with some text that clearly indicates this is not real data.

src/components/prescription-details.component.tsx
pirupius: Understood. Will rectify this in subsequent PRs

jonathandick:
@pirupius could you possible update the CI so that this deploys to dev3?

pirupius:
Sure thing @jonathandick

PR #13, opened 2022-09-14
O3-1501: Support Orders without an associated Drug

mogoodrich:
O3-1502: Format Date Time in Standard format
O3-1520: Remove (comment out?) "Cards" at top of the Dispensing page

### DIFF
--- a/src/components/medication-card.component.tsx
+++ b/src/components/medication-card.component.tsx
@@ -17,7 +17,11 @@ const MedicationCard: React.FC<{ medication: MedicationRequest }> = ({
     <Tile className={styles.medicationTile}>
       <div>
         <p className={styles.medicationName}>
-          <strong>{medication.medicationReference?.display}</strong>
+          <strong>
+            {medication.medicationReference
+              ? medication.medicationReference.display
+              : medication?.medicationCodeableConcept.text}
+          </strong>
         </p>
         <p className={styles.bodyLong01}>
           <span className={styles.label01}>
--- a/src/dispensing.component.tsx
+++ b/src/dispensing.component.tsx
@@ -7,7 +7,7 @@ export default function Dispensing() {
   return (
     <div className={`omrs-main-content`}>
       <PharmacyHeader />
-      <DispensingTiles />
+      {/*      <DispensingTiles />*/}
       <PrescriptionTabLists />
     </div>
   );
--- a/src/medication-request/medication-request.resource.tsx
+++ b/src/medication-request/medication-request.resource.tsx
@@ -1,66 +1,13 @@
 import useSWR from "swr";
-import {
-  fhirBaseUrl,
-  openmrsFetch,
-  openmrsObservableFetch,
-} from "@openmrs/esm-framework";
+import { fhirBaseUrl, openmrsFetch } from "@openmrs/esm-framework";
 import {
   AllergyIntoleranceResponse,
+  EncounterWithMedicationRequests,
+  EncountersWithMedicationRequestsResponse,
   MedicationRequest,
   MedicationRequestResponse,
+  Prescription,
 } from "../types";
-
-export type Order = {
-  id: string;
-  created: string;
-  patientName: string;
-  prescriber: string;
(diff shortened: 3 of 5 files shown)


src/components/medication-card.component.tsx
```diff
@@ -17,7 +17,11 @@ const MedicationCard: React.FC<{ medication: MedicationRequest }> = ({
     <Tile className={styles.medicationTile}>
       <div>
         <p className={styles.medicationName}>
-          <strong>{medication.medicationReference?.display}</strong>
+          <strong>
+            {medication.medicationReference
+              ? medication.medicationReference.display
+              : medication?.medicationCodeableConcept.text}
+          </strong>
```
mogoodrich, line 24: Updates the card to display requests with non-coded drugs

src/dispensing.component.tsx
```diff
@@ -7,7 +7,7 @@ export default function Dispensing() {
   return (
     <div className={`omrs-main-content`}>
       <PharmacyHeader />
-      <DispensingTiles />
+      {/*      <DispensingTiles />*/}
```
mogoodrich, line 10: Just commenting this out this I don't think these tabs will be implemented for v1 and I wanted to avoid having people reviewing the UI getting confused by them until they are implemented.

src/medication-request/medication-request.resource.tsx
```diff
@@ -1,79 +1,26 @@
...
-} from "@openmrs/esm-framework";
+import { fhirBaseUrl, openmrsFetch } from "@openmrs/esm-framework";
 import {
   AllergyIntoleranceResponse,
+  EncounterWithMedicationRequests,
+  EncountersWithMedicationRequestsResponse,
   MedicationRequest,
   MedicationRequestResponse,
+  Prescription,
 } from "../types";
```
mogoodrich, line 11: I extracted all the interfaces below in the "types" file and also renamed them to names that I thought were clearer... but feel free to weigh in if you disagree with the renaming.

src/medication-request/medication-request.resource.tsx
```diff
@@ -109,16 +57,22 @@ export function useOrders(
...
     patientName: encounter?.subject?.display,
-    drugs: [...new Set(orders.map((o) => o.medicationReference.display))].join(
-      ", "
-    ),
+    drugs: [
+      ...new Set(
+        orders.map((o) =>
+          o.medicationReference
+            ? o.medicationReference.display
+            : o.medicationCodeableConcept?.text
```
mogoodrich, line 72: Fix so that non-coded medications appear in the prescription list

src/prescriptions/prescription-tab-lists.component.tsx
```diff
@@ -154,7 +153,11 @@ const PrescriptionTabLists: React.FC = () => {
                                   <TableExpandRow {...getRowProps({ row })}>
                                     {row.cells.map((cell) => (
                                       <TableCell key={cell.id}>
-                                        {cell.value}
+                                        {cell.id.endsWith("created")
+                                          ? formatDatetime(
+                                              parseDate(cell.value)
+                                            )
+                                          : cell.value}
```
mogoodrich, line 160: I did the formatting here so that the underlying raw ISO string date value is used for sorting (and not the formatted day, which resulted in things being sorted alphabetically)... ideally, I think we should be parsing into a Date object so that that the Cell value would be a Date, not a string containing the raw ISO value) but this seemed to break the sort of the DataTable.

mseaton (approved):
LGTM, but good to get others to review

pirupius (approved):

src/prescriptions/prescription-tab-lists.component.tsx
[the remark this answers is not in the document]
mogoodrich: Thanks @pirupius ! The issue I was having was that if I just manipulated it in the resource then if someone clicks the sort arrows in the "Created" column, it sorts the dates alphabetically, not by date.

PR #15, opened 2022-09-23
O3-1218 create dispensation by filling form

pirupius:
The PR 

- adds dispense buttons on expanding list
- Launches dispense form
- successfully saves medication dispense to backend
- ui tweaks to list view


src/medication-dispense/medication-dispense.resource.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+export function initiateMedicationDispenseBody(
+  medicationRequests: Array<MedicationRequest>,
+  session: Session
+) {
+  let dispenseBody = [];
+  medicationRequests.map((medicationRequest) => {
+    let dispense = {
+      resourceType: "MedicationDispense",
+      status: "on-hold", // might need to change this to appropriate status
```
mogoodrich, line 41: yeah, I think this will need to be a coding reference (uuid or concept mapping) to an OpenMRS Concept in order to work... but, I don't think we defined exactly how we want this to work yet and I think it's fine to ticket this as a future enhancement.  Probably better to just not set the status at all for now, as I don't think "on-hold" will do anything/work?

src/medication-dispense/medication-dispense.resource.tsx
```diff
@@ -0,0 +1,85 @@
...
+      medicationReference: medicationRequest.medicationReference,
+      subject: medicationRequest.subject,
+      performer: [
+        {
+          actor: {
+            reference: medicationRequest.requester.reference,
```
mogoodrich, line 47: This shouldn't default to the requester, as the the person who dispensing the medication will almost certainly be someone different than the person prescribing it... this would likely default to the provider associated with the current user session, if that's available.  Otherwise could be left empty for now.

src/medication-dispense/medication-dispense.resource.tsx
```diff
@@ -0,0 +1,85 @@
...
+          },
+        },
+      ],
+      // location: {
+      //   reference: session?.sessionLocation?.uuid,
+      // },
```
mogoodrich, line 53: I do think this would be the session location, but if for some reason that is not currently working fine to keep it commented out for now.

src/medication-dispense/medication-dispense.resource.tsx
```diff
@@ -0,0 +1,85 @@
...
+      type: {
+        coding: [
+          {
+            code: "04affd1a-49ab-44e5-a6d1-c0a3fffceb7d", // what is this?
```
mogoodrich, line 57: This can likely be left blank for now, like the "status", we will likely want to build some logic around this, but I don't think we have figured that out yet.  Where did this uuid come from?

mogoodrich (approved):
Thanks @pirupius ! Looks good to me.  I didn't review all the components in great detail because I don't consider myself an expert with React, and it will likely be easier for me to review the UI elements once it's merged in and I can see them in action.

I did make a few comments on the default values for the medication dispense.  Let me know if they make sense.  I'm happy to get this merged in sooner rather than later.

Take care!

src/medication-dispense/medication-dispense.resource.tsx
pirupius: The documentation had this as a string along and it seems to be related to the ones defined in the list Tabs. We had only configured `on-hold` for testing purposes. This value is also related to the ongoing talk thread regarding the what statuses to support



> "[status](https://www.hl7.org/fhir/medicationdispense-definitions.html#MedicationDispense.status)" : "<[code](https://www.hl7.org/fhir/datatypes.html#code)>", // R!  [preparation | in-progress | cancelled | on-hold | completed | entered-in-error | stopped | declined | unknown](https://www.hl7.org/fhir/valueset-medicationdispense-status.html)

src/medication-dispense/medication-dispense.resource.tsx
pirupius: This was an oversight on my end.

src/medication-dispense/medication-dispense.resource.tsx
pirupius: The docs has it defined as this 👇🏾  but this looks like something that needs further discussion (definitely not MVP)

> "[type](https://www.hl7.org/fhir/medicationdispense-definitions.html#MedicationDispense.type)" : { [CodeableConcept](https://www.hl7.org/fhir/datatypes.html#CodeableConcept) }, // [Trial fill, partial fill, emergency fill, etc.](http://terminology.hl7.org/3.1.0/ValueSet-v3-ActPharmacySupplyType.html)

mogoodrich:
Thanks, all sounds good to me @pirupius !

PR #19, opened 2022-09-27
(chore) Trigger Bamboo build on merged commits

ibacher:
This should be harmless, but the idea is to trigger a build of the RefApp on each commit to the repo so that dev3 stays up-to-date with this module.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: "16"
+          node-version: "18"
       - run: yarn
       - run: yarn lint
       - run: yarn coverage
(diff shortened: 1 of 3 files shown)


.github/workflows/node.js.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: "16"
+          node-version: "18"
```
mogoodrich, line 25: probably makes sense to upgrade to Node 18 if that's what we are currently building with, just wanted to highlight that we are doing this here

mogoodrich (approved):
Thanks @ibacher !  Generally defer to you on this stuff....

pirupius (approved):

.github/workflows/node.js.yml
ibacher: The goal is to keep things compatible with the latest LTS and I think that switches over from 16 -> 18 in the next week or two. Whether that's sustainable long-term we'll have to see (there's some work in the pipeline to merge all of the MFs into a single monorepo which should make these tooling changes more straight-forward).

PR #21, opened 2022-10-07
change status when initially submitting dispense from "on-hold" to "c…

mogoodrich:
…omplete"

### DIFF
--- a/src/medication-dispense/medication-dispense.resource.tsx
+++ b/src/medication-dispense/medication-dispense.resource.tsx
@@ -38,7 +38,7 @@ export function initiateMedicationDispenseBody(
   medicationRequests.map((medicationRequest) => {
     let dispense = {
       resourceType: "MedicationDispense",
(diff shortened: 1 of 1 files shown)


src/medication-dispense/medication-dispense.resource.tsx
```diff
@@ -38,7 +38,7 @@ export function initiateMedicationDispenseBody(
   medicationRequests.map((medicationRequest) => {
     let dispense = {
       resourceType: "MedicationDispense",
-      status: "on-hold", // might need to change this to appropriate status
+      status: "completed", // might need to change this to appropriate status
```
mogoodrich, line 41: We still may need to tweak this, but "completed" is more accurate than "on-hold", so just updating this while it's on my mind.

pirupius: I agree with this.

pirupius (approved):

PR #22, opened 2022-11-02
O3-1571: Add Order and Dispense History to History Tab

mogoodrich:


src/medication-request/medication-request.resource.tsx
```diff
@@ -1,28 +1,24 @@
...
-  openmrsObservableFetch,
-  parseDate,
-} from "@openmrs/esm-framework";
+import { fhirBaseUrl, openmrsFetch, parseDate } from "@openmrs/esm-framework";
 import {
   AllergyIntoleranceResponse,
-  EncounterWithMedicationRequests,
-  EncountersWithMedicationRequestsResponse,
+  EncounterWithMedicationRequestsAndMedicationDispenses,
+  EncountersWithMedicationRequestsAndMedicationDispensesResponse,
```
pirupius, line 6: Names are getting longer 😅 

My only worry about this is that we might be coupling different types/tasks to to a single point. This isn't a blocker but definitely something that might need to be refactored later on.

pirupius (approved):

mseaton (approved):
Approving, but my review really isn't meaningful.

src/medication-request/medication-request.resource.tsx
mogoodrich: Yes, agreed @pirupius !  I actually wanted to start potentially refactoring the types into something clearer, but I figured I'd hold off a little bit while I learned more about how the code works, but you will likely see a renaming PR from me in the coming weeks!

mogoodrich:
Thanks @mseaton  and @pirupius !  This PR requires the changes to the FHIR module that have not yet been merged in, so I will wait on that before merging.

PR #24, opened 2022-11-04
refactor types: alphabetize, update responses to use resource types i…

mogoodrich:
…nstead of duplicating fields, rename "orders" to "medicationRequests"; rename "EncounterOrders" to prescription table

@pirupius  this is the beginning of some of the refactoring was looking to do, decided it was best to take it in small chunks.  Key things I did:

* sorted all the types in types.ts alphabetically
* removed some types that didn't appear to be used
* simplified the "Response" types so that they used the Encounter, MedicationRequest and MedicationDispense types instead of duplicating fields
* tried to remove the term "order" and replace with "medicationrequest" in an attempt for better clarity ("order" is the name in OpenMRS, "medication request" is the name in FHIR, seems best to try to stick to using one or the other)
* tried to give the type that represents a row in the main table a better name (renamed from EncounterOrder to PresciptionsTableRow... I'm not in love with that one yet, but it seemed like an improvement)

Hope to do some more refactoring next week, but this was the start!

### DIFF
--- a/src/medication-request/medication-request.resource.tsx
+++ b/src/medication-request/medication-request.resource.tsx
@@ -2,87 +2,87 @@ import useSWR from "swr";
 import { fhirBaseUrl, openmrsFetch, parseDate } from "@openmrs/esm-framework";
 import {
   AllergyIntoleranceResponse,
-  EncounterWithMedicationRequestsAndMedicationDispenses,
-  EncountersWithMedicationRequestsAndMedicationDispensesResponse,
+  EncounterResponse,
   MedicationRequest,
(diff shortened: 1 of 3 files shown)


mseaton (approved):
Good as far as I am concerned @mogoodrich .  Defer to @pirupius  on the specifics.

mogoodrich:
@pirupius  would be great for you to do a post-commit review... however, I'm going to merge this in now because I have further commits to apply on top of it.

pirupius (commented):
Looks good as far as i see.

PR #31, opened 2022-11-22
O3-1213: Improve Support for Status of Orders

mogoodrich:


mogoodrich:
@pirupius  @mseaton  this certainly not the final status calculation, but since we are starting to do more testing in the system, I wanted to put something in that was at least more sensical than the current logic (that just takes the status of the first element in the list).  This should display a proper status based on the 3 possible statuses that currently are returned by the API ("active", "cancelled", and "unknown")

pirupius (approved):
This is a good point that we can build on top. This could work for the mean time as we figure out how to best handle `MedicationDispense` entries

mogoodrich:
Thanks @pirupius , I'll merge this one in!

PR #32, opened 2022-11-22
rename order-expanded component to prescription-expanded component

mogoodrich:
refactor mutate of prescription details
add mutate of prescription table


src/components/prescription-expanded.component.tsx
```diff
@@ -3,7 +3,7 @@ import { PatientUuid } from "@openmrs/esm-framework";
 import { Tab, Tabs, TabList, TabPanels, TabPanel, Button } from "@carbon/react";
 import { useTranslation } from "react-i18next";
 import HistoryAndComments from "./history-and-comments.component";
-import styles from "./order-expanded.scss";
```
mogoodrich, line 6: I renamed "order-expanded" to "prescription-expanded" because it seems clearer with our current data model

src/medication-request/medication-request.resource.tsx
```diff
@@ -19,7 +19,7 @@ export function usePrescriptionsTable(
   pageOffset: number = 0,
   patientSearchTerm: string = ""
 ) {
-  const { data, error } = useSWR<{ data: EncounterResponse }, Error>(
+  const { data, mutate, error } = useSWR<{ data: EncounterResponse }, Error>(
```
mogoodrich, line 22: I created a mutate element at this level so we can update the prescription table row when a dispense form is saved

src/prescriptions/prescription-tab-lists.component.tsx
```diff
@@ -167,11 +167,12 @@ const PrescriptionTabLists: React.FC = () => {
                                   <TableExpandedRow
                                     colSpan={headers.length + 1}
                                   >
-                                    <OrderExpanded
+                                    <PrescriptionExpanded
                                       encounterUuid={row.id}
                                       patientUuid={
                                         encounterToPatientMap[row.id]
                                       }
+                                      mutatePrescriptionTableRows={mutate}
```
mogoodrich, line 175: I had to pass this mutate function from this tab list component, into the prescription expanded component, and then into the dispense form component... it seemed a little "ugly" to have to pass this mutate function around so much, if we have to do this further we might want to refactor to use a context.  @pirupius  I'm interested in your thoughts about the best way to do this.  The issue is we generally don't have access to the current page, nor the search string, of the rows currently being displayed.

src/forms/dispense-form.component.tsx
```diff
@@ -18,24 +17,23 @@ import {
...
   patientUuid,
   encounterUuid,
+  mutatePrescriptionTableRows,
 }) => {
   const { t } = useTranslation();
   const isTablet = useLayoutType() === "tablet";
   const session = useSession();
-  const { requests, isError, isLoading } =
-    usePrescriptionDetails(encounterUuid);
+  const { requests, mutate, isLoading } = usePrescriptionDetails(encounterUuid);
```
mogoodrich, line 35: I realized instead of pulling the generic mutate function off the SWFConfig, I could just generic bound mutate, which seemed cleaner.

src/prescriptions/prescription-tab-lists.component.tsx
pirupius: context might be a better alternative but this is something i might need to first pull down and test.

pirupius (approved):

pirupius:
@mogoodrich i've just noticed the base for this was the branch that has already been merged. This might need to be rebased before being able to merge in

src/prescriptions/prescription-tab-lists.component.tsx
mogoodrich: yeah, I didn't want to go overboard and add in a context if it wasn't really needed, as I have little experience with it.

PR #35, opened 2022-12-01
remove drug header

mogoodrich:
Just a tiny PR here... we are starting to demo to stakeholders and there was a request from the BAs to remove/hide the "1. Drug" header until we implement the "2. Internal Comments" header because it was confusing to have a "1" without a "2".


pirupius (approved):
👍🏾

PR #37, opened 2022-12-05
Updates tooling to Yarn 3 and adds a version export to the code

ibacher:
This is a pretty straight-forward update to use Yarn 3. It shouldn't require any changes in developer workflow in general.

This also bumps the versions of OpenMRS core depended on and adds a version number export.


ibacher:
Will look into this more...

PR #38, opened 2022-12-06
Refactors dispensing dashboard to use extensions for reuse

pirupius:
- Refactors dispensing dashboard to use extensions for reuse 
- Changes to dashboard tabs to show full text labels

Screenshot 2022-12-06 at 18 54 32

### DIFF
--- a/src/prescriptions/prescription-tab-lists.component.tsx
+++ b/src/prescriptions/prescription-tab-lists.component.tsx
@@ -12,15 +12,23 @@ enum TabTypes {
   ALL,
 }
 
-const tabs = [
-  { label: "activePrescriptions", status: "ACTIVE" },
-  { label: "allPrescriptions", status: "" },
-];
-
 const PrescriptionTabLists: React.FC = () => {
   const { t } = useTranslation();
   const [selectedTab, setSelectedTab] = useState(TabTypes.STARRED);
   const [searchTerm, setSearchTerm] = useState("");
+
+  const tabs = [
+    {
+      key: "activePrescriptions",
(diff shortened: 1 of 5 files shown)


src/prescriptions/prescription-tab-lists.component.tsx
```diff
@@ -39,12 +47,12 @@ const PrescriptionTabLists: React.FC = () => {
             {tabs.map((tab, index) => {
               return (
                 <Tab
-                  title={t(tab.label)}
+                  title={t(tab.key)}
                   key={index}
```
mogoodrich, line 51: Would it make sense to use the key and the key here, now that we have one?

src/prescriptions/prescriptions.scss
```diff
@@ -51,6 +51,10 @@ title {
   :global(.cds--tabs__nav-item--selected) {
     box-shadow: inset 0 2px 0 0 var(--brand-03) !important;
   }
+
+  :global(.cds--tab--list) button {
+    max-width: 12rem !important;
+  }
```
mogoodrich, line 57: Thanks! I noticed the tabs were cut off, but wasn't sure of where to tweak it!

mogoodrich (approved):
LGTM, thanks @pirupius , feel free to merge!

Just so I understand, what is the use case for the extension slot?

src/prescriptions/prescription-tab-lists.component.tsx
pirupius: Yes, this makes sense

PR #39, opened 2022-12-14
fix all styles to use branding variables instead of hardcoded colors

mogoodrich:


mogoodrich:
Thanks @mseaton !  @pirupius  this seems straightforward enough that I'm going to go ahead and merge, but post-commit reviews welcome...

pirupius:
@mogoodrich yes it is. Everything looks good.

PR #42, opened 2023-01-05
O3-1604 Allow changing Drug Formulation when dispensing

mogoodrich:


src/utils.ts
```diff
@@ -92,3 +95,34 @@ export function getPrescriptionTableEndpoint(
...
+export function getMedicationsByConceptEndpoint(conceptUuid: string) {
+  return `${fhirBaseUrl}/Medication?code=${conceptUuid}`;
+}
+
+export function getConceptUuidCoding(codings: Coding[]) {
+  // the concept uuid code is always the one without a system
+  return codings ? codings.find((c) => !("system" in c))?.code : null;
+}
+
+export function getOpenMRSMedicineDrugName(medication: Medication) {
```
mogoodrich, line 115: @ibacher  I based this on how the OpenMRS drug name extension appeared to be populated in the FHIR2 module for medications, appreciate you taking a look to see if there are any flaws in my logic or any reason this isn't a standard I should be relying on.  (It's working in my testing).

src/utils.ts
```diff
@@ -92,3 +95,42 @@ export function getPrescriptionTableEndpoint(
...
 }
+
+export function getMedicationsByConceptEndpoint(conceptUuid: string) {
+  return `${fhirBaseUrl}/Medication?code=${conceptUuid}`;
+}
+
+/**
+ * Given an array of CodeableConcept condings, return the first one without an associated system (which should be the uuid of the underyling concept)
+ * @param codings
+ */
```
mogoodrich, line 106: @ibacher  this is how it works currently, but thoughts on whether this is a standard that should be relied upon?

mseaton (approved):
Looks good to me, assuming that all of this is correctly interpreting the FHIR model (and concept and drug are populated either/or rather than both), etc.  I don't recall, but defer to @ibacher overall for review of FHIR-based assumptions.

mogoodrich:
Thanks @ibacher  and @mseaton , merging... @pirupius  as always, post-merge comments welcome... I'm still new to the ESMs and could be doing things incorrectly!

src/utils.ts
ibacher: Currently, this is the behaviour of the FHIR2 module, i.e., the only coding returned without a code system will be the UUID for the concept. This won't change without (at least) a major version of the FHIR2 module.

This is a little bit non-standard. So, let me explain a bit why we have this implementation (this is not germane to this PR, just describing general behaviour of the the FHIR2 module). In an ideal world, FHIR wants a code system for every code value (since without a system a code is ambiguous). As I see it, leaves us with three choices:

1. The current implementation (omit a system for codes that represent UUIDs; codes without systems are, as I understand it, implicitly understood to be implementation-specific, i.e., not really "interoperable" but potentially locally useful).
2. Create a dummy "OpenMRS Concept UUID" system and use that.
3. Create an implementation-specific system URL for, e.g., "Mirebalais Concept UUID" or "PIH Concept UUID".

Option 3 is the most semantically correct way to do things, but I don't see a good way to work out the right URL (basically, how do I know what implementation the current system is running?). 

Option 2 would violate FHIR's concept semantics because, for example, http://fhir.openmrs.org/cs/concept-uuid|1066AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA might have a different meaning inside PIH's dictionary than in does in, e.g., Uganda EMRs. The general problem here is concepts derived from CIEL that reuse the CIEL UUIDs but which have been customised in implementation-specific ways may no longer be semantically equivalent—and we have no way of catching.

So we ended up implementing option 1.

PR #45, opened 2023-01-09
Add translations

mogoodrich:


mogoodrich:
Just went through the code and:
* Looked for any raw text in the code
* Looked for any translation keys in the code that were missing in en.json

I'm sure I missed a couple, but hopefully I got most of them.

@pirupius  it would definitely still be great to set something up to do this automatically if you were interested, but I figured I'd just crank this out quickly as a first pass.

mseaton (approved):

pirupius (approved):
Thanks @mogoodrich for this. We'll just keep filling in the missing ones as we go on. 🦾 💯

PR #49, opened 2023-01-20
O3-1710: History and Comments Section: Ordered vs Dispensed

mogoodrich:
@pirupius  I realized after I was almost through this that there was some significiant refactoring that may affect some outstanding prescription code changes you may have, let me know if you run into issues.

A quick overview of what I did:

* Due requirements for different left-hand colors and different tags depending on whether in the prescription view and the history and comments view I extracted the "Tile" top-level element from the Medication Event Card component and renamed that Medication Event (because I think a "card" would reference the the whole element including the Tile, but let me know if you disagree with my naming convention).

* I also started moving things around a bit into folders by section, hopefully this makes things more organized, but it's a bit of a work in progress.


src/components/medication-event-card.component.tsx
```diff
@@ -1,98 +0,0 @@
...
-  const dosageInstruction: DosageInstruction = getDosageInstruction(
-    medicationEvent.dosageInstruction
-  );
-  const quantity: Quantity = getQuantity(medicationEvent);
-  const refillsAllowed: number = getRefillsAllowed(medicationEvent);
-
-  return (
-    <Tile className={styles.medicationEventTile}>
-      {actionMenu}
-      <div>
```
mogoodrich, line 35: Ugh, because I renamed, looks like we aren't getting a diff from Git, but rather it looks like the file has been deleted and created.  The main different in that the new "MedicationEvent" conmponent starts with this "div" here... the "Tile" and the "actionMenu" have been pulled out into the history-and-commments and prescription-details components.

src/history/history-and-comments.component.tsx
```diff
@@ -87,10 +89,11 @@ const HistoryAndComments: React.FC<{
...
               </h5>
-              <MedicationEventCard
-                medicationEvent={dispense}
-                actionMenu={generateMedicationDispenseActionMenu(dispense)}
-              />
+              <Tile className={styles.dispenseTile}>
+                {generateMedicationDispenseActionMenu(dispense)}
+                <Tag type="red">{t("dispense", "Dispense")}</Tag>
+                <MedicationEvent medicationEvent={dispense} />
+              </Tile>
```
mogoodrich, line 96: This is a clearer example of the what changed in the MedicationEventCard.... the Tile now leaves here (so we can apply custom styling) as well the action menu, and the new "Dispense" tag... the Medication Event card just displays the details of the medication event.

src/prescriptions/prescription-details.component.tsx
```diff
@@ -59,7 +79,13 @@ const PrescriptionDetails: React.FC<{
       {requests &&
         requests.map((request) => {
           return (
-            <MedicationEventCard key={request.id} medicationEvent={request} />
+            <Tile className={styles.prescriptionTile}>
+              <MedicationEvent
```
mogoodrich, line 83: Same here... the Tile has been pulled out into the Prescription Details component so that custom left-hand strip can be applied.

src/prescriptions/prescription-details.component.tsx
```diff
@@ -26,6 +27,25 @@ const PrescriptionDetails: React.FC<{
...
+      return <Tag type="red">{t("expired", "Expired")}</Tag>;
+    }
+
+    if (medicationRequest.status === "cancelled") {
+      return <Tag type="red">{t("cancelled", "Cancelled")}</Tag>;
+    }
+    // TODO support completed status & will need to change expired once we change the definition of expired
+    // TODO will need to support potential Medication Dispense statuses or refactor into different request and dispense components
+    return null;
+  };
```
mogoodrich, line 47: I pulled this out of a separate tag into this details component because it is now the only place where it is being used.  It remains to be seen if we'll want to pull this out again.

mseaton (approved):
Approving, but defer to @pirupius to confirm it looks ok

pirupius (approved):
As usual love the comments 🤩 and i have no objections with the renaming and the restructure was bound to happen as the codebase grew. Please go ahead with the merge.

mogoodrich:
thanks @pirupius , merging!

PR #50, opened 2023-02-03
O3-1608: Set "Reason for Change" when substituting drugs

mogoodrich:
Note that there is still a slight problem is that we can't "clear out" the change reasons when editing an existing dispense event, see:

https://talk.openmrs.org/t/setting-property-values-to-null-when-submitting-fhir-objects-via-fhir2-module/38471/3

### DIFF
--- a/src/components/medication-dispense-review.component.tsx
+++ b/src/components/medication-dispense-review.component.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { CommonConfigProps, Medication, MedicationDispense } from "../types";
 import MedicationCard from "./medication-card.component";
 import { TextArea, ComboBox, Dropdown, NumberInput } from "@carbon/react";
@@ -14,6 +14,7 @@ import {
   useMedicationCodeableConcept,
   useMedicationFormulations,
 } from "../medication/medication.resource";
+import { useMedicationRequest } from "../medication-request/medication-request.resource";
 
 interface MedicationDispenseReviewProps {
   medicationDispense: MedicationDispense;
@@ -23,6 +24,8 @@ interface MedicationDispenseReviewProps {
   drugDispensingUnits: Array<CommonConfigProps>;
   drugRoutes: Array<CommonConfigProps>;
   orderFrequencies: Array<CommonConfigProps>;
+  substitutionReasons: Array<CommonConfigProps>;
+  substitutionTypes: Array<CommonConfigProps>;
 }
 
 const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
@@ -33,9 +36,12 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
   drugDispensingUnits,
   drugRoutes,
   orderFrequencies,
+  substitutionReasons,
+  substitutionTypes,
 }) => {
   const { t } = useTranslation();
   const [isEditingFormulation, setIsEditingFormulation] = useState(false);
+  const [isSubstitution, setIsSubstitution] = useState(false);
   const isTablet = useLayoutType() === "tablet";
 
   // if this an order for a drug by concept, but not a particular formulation, we already have access to concept uuid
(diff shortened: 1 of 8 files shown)


src/components/medication-dispense-review.component.tsx
```diff
@@ -14,30 +14,135 @@ import {
...
   const [isEditingFormulation, setIsEditingFormulation] = useState(false);
+  const [isSubstitution, setIsSubstitution] = useState(false);
+  // Dosing Unit eg Tablets
+  const [drugDosingUnits, setDrugDosingUnits] = useState([]);
+  // Dispensing Unit eg Tablets
+  const [drugDispensingUnits, setDrugDispensingUnits] = useState([]);
+  // Route eg Oral
+  const [drugRoutes, setDrugRoutes] = useState([]);
+  // Frequency eg Twice daily
+  const [orderFrequencies, setOrderFrequencies] = useState([]);
```
mogoodrich, line 47: When adding the new substitution options for the two new substitution dropdowns, I realized that we could pull all the config stuff down into the component... I think SWR caching will make sure it doesn't re-fetch the options each time?  

This saves passing all this stuff down as props from the dispense-form component, but let me know if I've got something wrong @pirupius

src/components/medication-dispense-review.component.tsx
```diff
@@ -14,30 +14,135 @@ import {
...
+      orderConfigObject.orderFrequencies.forEach(
+        (x) =>
+          availableFrequencies.includes(x.uuid) ||
+          otherFrequencyOptions.push({ id: x.uuid, text: x.display })
+      );
+      setOrderFrequencies([...orderFrequencies, ...otherFrequencyOptions]);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [orderConfigObject]);
+
```
mogoodrich, line 107: Above effects were just moved from dispense-form, new effects to load substitution stuff below.

src/components/medication-dispense-review.component.tsx
```diff
@@ -59,6 +164,59 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
       : null
   );
 
+  // get the medication request associated with this dispense event
+  // (we fetch this so that we can use it below to determine if the formulation dispensed varies from what was
+  // ordered; it's slightly inefficient/awkward to fetch it from the server here because we *have* fetched it earlier,
+  // it just seems cleaner to fetch it here rather than to make sure we pass it down through various components; and
+  // SWR should handle caching properly)
```
mogoodrich: fyi, see my long-winded comment above.. :)

src/forms/dispense-form.component.tsx
```diff
@@ -145,51 +136,6 @@ const DispenseForm: React.FC<DispenseFormProps> = ({
 
   useEffect(checkIsValid, [medicationDispensesPayload]);
 
-  useEffect(() => {
```
mogoodrich, line 148: see my comments on the medication-dispense-review component

src/medication-dispense/medication-dispense.resource.tsx
```diff
@@ -117,6 +142,12 @@ export function initiateMedicationDispenseBody(
         ],
         substitution: {
           wasSubstituted: false,
+          reason: [
+            {
+              coding: [{ code: null }],
+            },
+          ],
+          type: { coding: [{ code: null }] },
```
mogoodrich, line 150: We will probably need to tweak this based on:

https://talk.openmrs.org/t/setting-property-values-to-null-when-submitting-fhir-objects-via-fhir2-module/38471/3

src/components/medication-dispense-review.component.tsx
```diff
@@ -59,6 +164,59 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
       : null
   );
 
+  // get the medication request associated with this dispense event
+  // (we fetch this so that we can use it below to determine if the formulation dispensed varies from what was
+  // ordered; it's slightly inefficient/awkward to fetch it from the server here because we *have* fetched it earlier,
+  // it just seems cleaner to fetch it here rather than to make sure we pass it down through various components; with
+  // SWR handling caching, we may want to consider pulling more down into this)
```
mogoodrich, line 171: See my long-winded comment above... :)

src/config-schema.ts
```diff
@@ -25,8 +25,30 @@ export const configSchema = {
     _type: Type.String,
     _default: "Pharmacy",
   },
+  substitutionReason: {
```
mogoodrich, line 28: Note these two new configuration parameters.  Will be required, or the dropdown for these two questions will be rendered empty.

mseaton (commented):
Thanks @mogoodrich .  Will defer most of the review to the others.  Let me know if you need me to take a closer look.

src/components/medication-dispense-review.component.tsx
pirupius: You have a point about the caching. The main worry initially was to avoid making the repeated calls.

pirupius (approved):
LGTM

src/components/medication-dispense-review.component.tsx
mogoodrich: Yeah, I *think* with SWR it handles caching for you, but we can refactor back if it doesn't.

It also got me thinking about other places where we *don't* want it to cache... but I've got a separate to-do to myself to ticket that

mogoodrich:
Thanks @pirupius !

PR #51, opened 2023-02-10
O3-1712: Analysis: Rework "Expired" in the context of dispensing

mogoodrich:
Note this is dependent on this FHIR PR being approved and merged in:
https://github.com/openmrs/openmrs-module-fhir2/pull/459

### DIFF
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -25,8 +25,15 @@ export const configSchema = {
     _type: Type.String,
     _default: "Pharmacy",
   },
+  medicationRequestExpirationPeriodInDays: {
+    _type: Type.Number,
+    _description:
+      "Medication Requests older that this will be considered expired",
+    _default: 90,
+  },
 };
 
 export type PharmacyConfig = {
   appName: String;
+  medicationRequestExpirationPeriodInDays;
 };
--- a/src/medication-request/medication-request.resource.tsx
+++ b/src/medication-request/medication-request.resource.tsx
@@ -131,7 +131,7 @@ function buildPrescriptionsTableRow(
     prescriber: [
       ...new Set(medicationRequests.map((o) => o.requester.display)),
     ].join(", "),
-    status: computeStatus(
+    status: computePrescriptionStatus(
       medicationRequests.map((o) => o.status),
       encounter?.period?.start,
       medicationRequestExpirationPeriodInDays
@@ -140,7 +140,7 @@ function buildPrescriptionsTableRow(
   };
 }
 
-function computeStatus(
+function computePrescriptionStatus(
   orderStatuses: Array<string>,
   encounterDatetime: string,
   medicationRequestExpirationPeriodInDays
--- a/src/medication-request/medication-request.resource.tsx
+++ b/src/medication-request/medication-request.resource.tsx
@@ -1,10 +1,5 @@
 import useSWR from "swr";
-import {
-  fhirBaseUrl,
-  openmrsFetch,
-  parseDate,
-  useConfig,
-} from "@openmrs/esm-framework";
+import { fhirBaseUrl, openmrsFetch, parseDate } from "@openmrs/esm-framework";
 import {
   AllergyIntoleranceResponse,
   EncounterResponse,
--- a/src/medication-request/medication-request.resource.tsx
+++ b/src/medication-request/medication-request.resource.tsx
@@ -31,11 +31,10 @@ export function usePrescriptionsTable(
           pageOffset,
           pageSize,
           patientSearchTerm,
-          "ge" +
-            dayjs(new Date())
-              .startOf("day")
-              .subtract(medicationRequestExpirationPeriodInDays, "day")
-              .toISOString()
+          dayjs(new Date())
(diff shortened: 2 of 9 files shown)


src/config-schema.ts
```diff
@@ -25,6 +25,12 @@ export const configSchema = {
     _type: Type.String,
     _default: "Pharmacy",
   },
+  medicationRequestExpirationPeriodInDays: {
+    _type: Type.Number,
+    _description:
+      "Medication Requests older that this will be considered expired",
+    _default: 90,
+  },
```
mogoodrich, line 33: New config parameter to configure the expiration period in days (which is compared against the encounter date)

src/medication-request/medication-request.resource.tsx
```diff
@@ -117,25 +131,63 @@ function buildPrescriptionsTableRow(
...
 }
 
-function computeStatus(orderStatuses: Array<string>) {
-  if (orderStatuses.includes("active")) {
-    // if any are active, return active
-    return "active";
-  } else if (orderStatuses.includes("stopped")) {
-    // if none are active, and any are stopped, return expired
-    return "expired";
+function computeStatus(
```
mogoodrich, line 143: Updated logic to determine the overall status of Medication Requests linked to an encounter...  this (should) mirror the logic in the encountersWithMedicationRequests named query provided by the FHIR module.

Hopefully, documentation makes sense.

src/utils.ts
```diff
@@ -87,13 +87,21 @@ export function getPrescriptionDetailsEndpoint(encounterUuid: string) {
...
-export function getPrescriptionTableEndpoint(
+export function getPrescriptionTableActiveMedicationRequestsEndpoint(
   pageOffset: number,
   pageSize: number,
   patientSearchTerm: string,
-  status: string
+  date: string
 ) {
-  return `${fhirBaseUrl}/Encounter?_getpagesoffset=${pageOffset}&_count=${pageSize}&subject.name=${patientSearchTerm}&_revinclude=MedicationRequest:encounter&_revinclude:iterate=MedicationDispense:prescription&_has:MedicationRequest:encounter:status=${status}&_sort=-date&_tag=http%3A%2F%2Ffhir.openmrs.org%2Fext%2Fencounter-tag%7Cencounter`;
+  return `${fhirBaseUrl}/Encounter?_query=encountersWithMedicationRequests&_getpagesoffset=${pageOffset}&_count=${pageSize}&patientSearchTerm=${patientSearchTerm}&date=${date}&status=active`;
```
mogoodrich, line 96: Note that we've switched this request and the one below to use the new named query "encountersWithMedicationRequests"

src/prescriptions/prescription-details.component.tsx
```diff
@@ -30,19 +33,27 @@ const PrescriptionDetails: React.FC<{
   const generateStatusTag: Function = (
     medicationRequest: MedicationRequest
   ) => {
-    if (!medicationRequest.status || medicationRequest.status === "active") {
-      return null;
+    if (medicationRequest.status === "cancelled") {
```
mogoodrich, line 36: updated logic to determine the status of an *individual* medication request

mogoodrich:
Ready for review, see my comment about not merging until the appropriate FHIR PR is approved and merged.

mseaton (approved):
From a code perspective it seems fine, I'm still not sure conceptually how much I like expiry logic as a frontend-application-level construct.

src/config-schema.ts
pirupius: Does this mean logic not to allow expired requests has been added as well? And also how does this work in relation to recurring medications?

src/medication-request/medication-request.resource.tsx
```diff
@@ -10,25 +10,37 @@ import {
...
-      status
-    ),
+    status === "ACTIVE"
+      ? getPrescriptionTableActiveMedicationRequestsEndpoint(
+          pageOffset,
+          pageSize,
+          patientSearchTerm,
+          dayjs(new Date())
+            .startOf("day")
+            .subtract(medicationRequestExpirationPeriodInDays, "day")
```
pirupius, line 36: 👍🏾

pirupius (approved):
Just have the one question in regards to how the request expiry related to recurring medications. Has that been put into consideration before being marked as expired? I see the main focus is on the encounter dateTime

src/config-schema.ts
mogoodrich: Yes, right now you can't dispense orders older than x days old, (which a default of 90)... you are right that this will likely need to be tweaked based on refills, see my comment below.

mogoodrich:
@pirupius regarding expire, yes, it's frustratingly overcomplicated... :)

We are awaiting discussions with our lab teams about how refills fit into the model of the expired, and it seems likely that the logic will need to be tweaks depending on refills, this is just a first past while we figure the out.  If the lab term on your end has any thoughts on this, feel free to have them reach out to Fiona.

Thanks for the review, I'll merge this in after the FHIR PR gets merged (which might not be until next week because I believe Ian is on vacation this week).

PR #52, opened 2023-02-21
"debounce" the search function so that it isn't triggered on every ke…

mogoodrich:
…ystroke

### DIFF
--- a/src/prescriptions/prescription-tab-lists.component.tsx
+++ b/src/prescriptions/prescription-tab-lists.component.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { Tab, Tabs, TabList, TabPanels, Search, Button } from "@carbon/react";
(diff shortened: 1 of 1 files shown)


src/prescriptions/prescription-tab-lists.component.tsx
```diff
@@ -30,6 +31,15 @@ const PrescriptionTabLists: React.FC = () => {
     },
   ];
 
+  // debounce: delay the search term update so that a search isn't triggered on every single keystroke
+  useEffect(() => {
+    const debounceFn = setTimeout(() => {
+      setSearchTerm(searchTermUserInput);
+    }, 500);
```
mogoodrich, line 38: We make sure we wait 500ms before updating the search term prop so that we don't trigger a new search on *every* keystroke

mogoodrich:
@pirupius  @mseaton  I'm going to go ahead and merge this so that I can test it on our staging servers, but post-commit reviews *definitely* welcome!

PR #53, opened 2023-02-21
O3-1709: Permissions for which Dispensing fields can be edited and by which role

mogoodrich:
See: https://issues.openmrs.org/browse/O3-1709

Note that if we agree these are the correct names, we will need to add these privileges to the reference app

I will also ticket and implement an update to the EMR-API so that privileges that start with "o3." are excluded when giving a user all API privileges.


mogoodrich:
Closing this for now... I had forgotten that I wanted to add these privileges as constants, which I haven't done yet... should re-issue tomorrow.

PR #54, opened 2023-02-22
O3-1709: Permissions for which Dispensing fields can be edited and by which role

mogoodrich:


src/prescriptions/prescription-expanded.component.tsx
```diff
@@ -67,24 +68,26 @@ const PrescriptionExpanded: React.FC<{
...
+                )
+              }
+            >
+              {t("dispense", "Dispense")}
+            </Button>
+          </UserHasAccess>
+          {/*  <Button kind="secondary" className={styles.returnToPrescriberBtn}>
             {t("sendBackToPrescriber", "Send back to prescriber")}
-          </Button>
+          </Button>*/}
```
mogoodrich, line 90: Note, tangentially, I commented this button out until we implement it.

mogoodrich (commented):
Okay, now this is ready for you... @ibacher would be good to get your thoughts about the naming convention I came up for the "o3" privileges.

I'll open another PR against EMR-API to make sure that we properly group these new "o3." privileges as "application level" privileges.

README.md
```diff
@@ -5,6 +5,16 @@
 This repository is for the OpenMRS Dispensing App. For more information, please see the
 [Design Documentation](https://wiki.openmrs.org/display/projects/Dispensing+Design+Components).
 
+## Required privileges
+
+Note that following privileges need to be installed and assigned to roles:
+
+* o3.dispensing-app.dispense.create: Allows user to Dispense Medication
+* o3.dispense-app.dispense.create.modifyDetails: Allows user to modify the Quantity, Drug, Formulation and Dose Instructions (from the values specified in the Order / Medication Request) when Dispensing
```
mseaton, line 13: dispense-app or dispensing-app?  Do we really need the o3 prefix?

README.md
```diff
@@ -5,6 +5,16 @@
...
+* o3.dispensing-app.dispense.edit: Allows user to edit an existing Medication Dispense
+* o3.dispensing-app.dispense.delete: Allows user to delete an existing Medication Dispense 
+* o3.dispensing.delete.thisProviderOnly: Allows user to delete an existing Medication Dispense, *but only* if they created it originally
```
mseaton, line 16: delete.ifCreator?

src/constants.ts
```diff
@@ -6,3 +6,11 @@ export const spaBasePath = `${window.spaBase}${basePath}`;
 export const OPENMRS_FHIR_PREFIX = "http://fhir.openmrs.org";
 export const OPENMRS_FHIR_EXT_PREFIX = OPENMRS_FHIR_PREFIX + "/ext";
 export const OPENMRS_FHIR_EXT_MEDICINE = OPENMRS_FHIR_EXT_PREFIX + "/medicine";
+
+export const PRIVILEGE_CREATE_DISPENSE = "o3.dispensing-app.dispense.create";
+export const PRIVILEGE_CREATE_DISPENSE_MODIFY_DETAILS =
```
mseaton, line 11: see comment above with inconsistency in wording

mseaton (changes requested):

README.md
mogoodrich: Oh, nice catch, I did this while waiting to board my red-eye and looks like I got sloppy and missed this, will correct.

I'll comment about the "o3" prefix on the other thread....

mogoodrich:
Updated to use "Task:" privileges instead of the new "o3" privileges, also make some tweaks to privilege names, ready for re-review @mseaton  @pirupius

README.md
```diff
@@ -5,6 +5,16 @@
 This repository is for the OpenMRS Dispensing App. For more information, please see the
 [Design Documentation](https://wiki.openmrs.org/display/projects/Dispensing+Design+Components).
 
+## Required privileges
+
+Note that following privileges need to be installed and assigned to roles:
+
+* "Task: dispensing.create.dispense" - Allows user to Dispense Medication
+* "Task: dispensing.create.dispense.andModifyDetails" - Allows user to modify the Quantity, Drug, Formulation and Dose Instructions (from the values specified in the Order / Medication Request) when Dispensing
```
mseaton, line 13: I don't love this name.  Isn't this really "dispensing.allow.substitutions" or something like that?

README.md
```diff
@@ -5,6 +5,16 @@
...
+* "Task: dispensing.edit.dispense" - Allows user to edit an existing Medication Dispense
```
mseaton, line 14: Is there a world where we would want this to be different than the "create" privileges?  Would we ever want to prevent someone from editing something they created?  What if they made a typo?

mseaton (approved):
Fine with me.  Made some comments in case they are helpful that you can take if you want.

README.md
mogoodrich: I do like that better, will tweak

mogoodrich:
Thanks @mseaton for the review, I'm going to go ahead and merge this in, but post-merge reviews welcome @pirupius  @ibacher 

Also: @ibacher  is there somewhere I should request adding these privileges to the O3 ref app? Do we need to worry about that?  Alos, @pirupius  note that these privileges will need to be added to the OHRI system as well?

ibacher:
> is there somewhere I should request adding these privileges to the O3 ref app?

It seems like a good idea to ensure the RefApp has all the privileges for apps that are part of it. For right now, I think just a PR to [this file](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/main/distro/configuration/privileges/privileges_core-demo.csv) is sufficient.

PR #55, opened 2023-02-25
O3-1919: Excessive re-rendering when opening dispensing form

mogoodrich:
See: https://issues.openmrs.org/browse/O3-1919


src/forms/initialize-dispense-form-from-requests.component.tsx
```diff
@@ -42,7 +42,11 @@ const InitializeDispenseFormFromRequests: React.FC<NewDispenseFormProps> = ({
       );
       setMedicationDispenses(dispenseMedications);
     }
-  }, [requests, session, config.medicationRequestExpirationPeriodInDays]);
+  }, [
+    requests.length,
+    session,
```
mogoodrich, line 47: only change is to compare requests.length instead of just requests

pirupius (approved):

PR #58, opened 2023-03-03
O3-1837: Add Location column to the Dispense app

mogoodrich:
@pirupius  this was a request from our implementations... I did set it up so the location column and filter only appears if an implementation has multiple login locations... seems like a useful feature for everyone, but let me know if this is something you think we need to be able to "turn off" for implementations that don't want it:


mseaton (commented):
I'll leave the React code to @pirupius to review, but one question I'd have is whether "Login Location" is the right construct for this (eg. and not Dispensing Location or Visit Location or some other Location tag).

mogoodrich:
@mseaton  it's not the dispensing location, it's the order location, and we don't have a order location tag, nor do I really think we want one as we aim to incorporate orders into the standard encounter workflow... basically, we are displaying all encounters with drug orders, and want to display/filter by encounter location.  It's also a server-side, paginated search, so we can't just filter by the locations in the result set.

src/location/location.resource.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }));
+    const loginLocations = useLoginLocations();
+    expect(useSWR).toHaveBeenCalledWith(
+      "/ws/fhir2/R4/Location?_tag=login%20location&_count=100",
+      openmrsFetch
+    );
+  });
+  test("useLoginLocations should parse into Login Locations Array", () => {
+    // @ts-ignore
+    const queryResultsBundle = {
```
pirupius, line 32: i think we can move this to separate mock file so that we don't have a lot of these in the test file

pirupius (approved):
Approved with a suggestion. Also i think this could be changed to a configuration and disabled by default.

src/config-schema.ts
```diff
@@ -31,6 +31,30 @@ export const configSchema = {
...
+        _default: false,
+      },
+      tag: {
+        _type: Type.String,
+        _description:
+          "Name of the location tag to use when fetching locations to populate filter",
+        _default: "Login Location",
+      },
+    },
+  },
```
mogoodrich, line 57: See above @mseaton @pirupius

mogoodrich (commented):
So @pirupius @mseaton  I added some configuration parameters based on the feedback to enable/disable the location column and the location filter (both default to false) and to set the tag to use when fetching locations (defaults to Login Location).  I flagged you on the config so you can take a look, opening for re-review.

Also @pirupius  I wasn't quite sure of the best way to add a mock for useSWR in a separate file... thoughts? (And can we handle that as a separate ticket?)

mseaton (commented):
LGTM, but defer to @pirupius

mogoodrich:
Thanks @mseaton !  Merging, but @pirupius  please take a look at my config changes when you get a chance and let me know if that meets the OHRI requirements, thanks!

PR #70, opened 2023-05-03
O3-2087: Support Status on Overall Prescription

mogoodrich:
## Requirements

### DIFF
--- a/src/components/action-buttons.component.tsx
+++ b/src/components/action-buttons.component.tsx
@@ -23,15 +23,15 @@ import CloseDispenseForm from "../forms/close-dispense-form.component";
 interface ActionButtonsProps {
   medicationRequest: MedicationRequest;
   associatedMedicationDispenses: Array<MedicationDispense>;
-  mutate: Function;
   patientUuid: string;
+  encounterUuid: string;
 }
 
 const ActionButtons: React.FC<ActionButtonsProps> = ({
   medicationRequest,
   associatedMedicationDispenses,
-  mutate,
   patientUuid,
+  encounterUuid,
 }) => {
   const { t } = useTranslation();
   const config = useConfig() as PharmacyConfig;
@@ -67,13 +67,13 @@ const ActionButtons: React.FC<ActionButtonsProps> = ({
               t("dispensePrescription", "Dispense prescription"),
               <DispenseForm
                 patientUuid={patientUuid}
+                encounterUuid={encounterUuid}
                 medicationDispense={initiateMedicationDispenseBody(
                   medicationRequest,
                   session,
                   true
                 )}
                 mode="enter"
-                mutate={mutate}
               />
             )
           }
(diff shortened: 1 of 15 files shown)


src/types.ts
```diff
@@ -350,17 +351,25 @@ export enum MedicationRequestStatus {
...
+export enum MedicationRequestFulfillerStatus {
+  on_hold = "ON_HOLD",
+  declined = "DECLINED",
+  completed = "COMPLETED",
+}
+
+/**
+ * Within the UI, the "status" of a request we want to display to the pharmacist is
+ * a combination of the status and fulfiller status; this is the actual status we want to display to the pharmacist (see util method
+ * computeMedicationRequestCombinedStatus)
```
mogoodrich, line 366: This is a rework, we since we are now updating the fulfiller status on a medication request, we now calculate the "combined" condition based on the status and fulfiller status of the medication request (and not the medication dispense status).... see comments above.

src/utils.ts
```diff
@@ -3,53 +3,76 @@ import {
...
 } from "./constants";
 import dayjs from "dayjs";
 
 /**
- * This is a bit of a mouthful, but, within the UI, the "status" of a request we want to display to the pharmacist is
- * a combination of the status of the request and the status of the most recent dispensing event; given a request
- * status and a dispense status, this calculates the actual status we want to display to the pharmacist
+ * Within the UI, the "status" of a request we want to display to the pharmacist is
+ * a combination of the status and the fulfiller statuts; given a request
+ * this calculates the actual status we want to display to the pharmacist
```
mogoodrich, line 24: here's the updated calculation for the "combined status" which I talked about in my comment above.

src/utils.ts
```diff
@@ -139,6 +274,16 @@ export function getDosageInstruction(
   return null;
 }
 
+/**
```
mogoodrich, line 277: The rest are all new utility methods that hopefully are understandable based on the comments.

mogoodrich (commented):
Ready for review!  

Note that these changes will need some new changes from the FHIR2 module to be merged in, so please review but don't merge in yet, thanks!

src/components/action-buttons.component.tsx
```diff
@@ -23,15 +23,15 @@ import CloseDispenseForm from "../forms/close-dispense-form.component";
 interface ActionButtonsProps {
   medicationRequest: MedicationRequest;
   associatedMedicationDispenses: Array<MedicationDispense>;
-  mutate: Function;
```
mogoodrich, line 26: I refactored the mutate functionality to use a global mutate function instead of a bound mutate, so the mutate functionality has all been extracted out into utils.tsx, see my comments there.

src/utils.ts
```diff
@@ -275,6 +436,52 @@ export function getRefillsAllowed(
...
+    sorted.length > 0 &&
+    sorted[0].id === medicationDispense.id ? true : false;
+}
+
+/**
+ * Revalidated (reloads) both the prescription associated with the encounter uuid,
+ * and the entire prescrption table
+ * @param encounterUuid
+ */
+export function revalidate(encounterUuid: string) {
```
mogoodrich, line 474: This utility function uses the useSWR mutate function to trigger a revalidation of the SWR, so that the data is updated on any change.

Each time we make a change we need to:
1) reload all tabs in the prescription table (as some prescriptions may be added/removed from a table due to state change)
2) reload the prescriptions details of the specific prescription/encounter we are working with

Thie method provides a filter function to match and invalidate the appropriate keys.

mogoodrich:
@pirupius @denniskigen  @ibacher  I'm going to go ahead and merge this in (I've got another PR based off it coming in shortly), but post-commit reviews definiteyl welcome!

PR #73, opened 2023-05-11
(chore)  O3-1514: Upgrade Yarn to berry

Jexsie:
## Requirements

## Summary

- Upgrades app to yarn berry

## Related Issue

- https://issues.openmrs.org/browse/O3-1514

## Other

- The workflow still remains the same ie `yarn` or `yarn install` for dependency installation


Jexsie:
cc @pirupius @mogoodrich

package.json
```diff
@@ -70,6 +68,7 @@
     "@types/webpack-env": "^1.16.0",
     "@typescript-eslint/parser": "^4.26.0",
     "concurrently": "^6.2.0",
+    "dayjs": "^1.11.7",
```
ibacher, line 71: Why did this get moved? (dayjs itself is tiny, but I'd rather just have one instance of it so we can all use one set of plugins).

pirupius, line 71: I agree with Ian on this. It should be moved back to the `dependences`

pirupius (changes requested):

package.json
Jexsie: We had issues where `dayjs` was not to be installed for development with `yarn` Moving it to the dev dependencies kinda solved that.

Jexsie: you mean `peer dependencies`

pirupius: My bad 😅  Yes `peerDependencies`

ibacher: It's fine to have it in both `devDependencies` and `peerDependencies`... that's our usual pattern. However, it definitely should be in `peerDepedencies` as otherwise we'll have runtime issues.

ibacher: Note that in the monorepos, the usual pattern is actually that it's in `peerDependencies` for the app and `devDependencies` for the repo packages.json, but for single-app packages, it will all be in the one place.

denniskigen:
@ibacher please review again.

mogoodrich:
I haven't been following this too closely, @Jexsie  do you have more changes to make, or are you waiting for  re-review from @denniskigen , @ibacher  and @pirupius ?

denniskigen:
Should be good to go, @mogoodrich, I'd asked @ibacher to review it anew.

ibacher (approved):
LGTM

mogoodrich:
Cool, thanks @denniskigen  and @ibacher ... @Jexsie  I'll test tomorrow to make sure things are working for me and then I will close the ticket out, thanks again!

PR #87, opened 2023-09-05
O3-2399: Dispensing: Set refresh interval so that table automatically refreshes

mogoodrich:
## Requirements

### DIFF
--- a/src/medication-request/medication-request.resource.tsx
+++ b/src/medication-request/medication-request.resource.tsx
@@ -27,7 +27,7 @@ import {
   OPENMRS_FHIR_EXT_REQUEST_FULFILLER_STATUS,
 } from "../constants";
 
-const REFRESH_INTERVAL = 10000; // 10 second refresh interval
+export const REFRESH_INTERVAL = 10000; // 10 second refresh interval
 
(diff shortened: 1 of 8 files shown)


src/medication-request/medication-request.resource.tsx
```diff
@@ -27,6 +27,8 @@ import {
   OPENMRS_FHIR_EXT_REQUEST_FULFILLER_STATUS,
 } from "../constants";
 
+export const REFRESH_INTERVAL = 10000; // 10 second refresh interval
```
mseaton, line 30: Seems fine.  If easy, could also pull this from config with a default of 10000, which would allow changing this if after testing a different value is desired.

mseaton (approved):
LGTM, just one comment

mogoodrich (commented):
Ready for re-review, updated to use a config parameter @mseaton  @pirupius 

As a side note, one thing I should look into:  a hook can only be used in a component, so I'm not able to access the config variable within resource itself, but basically have to pass it through from the components via a new parameter on the API methods.

mseaton (approved):

pirupius (approved):

PR #90, opened 2023-11-13
(feat) Use fewer Webpack chunks

ibacher:
## Requirements

## Summary

This just applies the pattern to ensure that most components are in the main Webpack chunk for the app, hopefully resulting in fewer overall JS file requests.


mogoodrich:
Sorry @ibacher  I haven't been following this whole thread, is there a relevant Talk post or ticket that has a summary? Thanks!

ibacher:
@mogoodrich I don't think so, it's mostly just been discussed on calls. The main idea here is that we have too many JS files, which is caused by Webpack's chunking algorithm. Any time we have a dynamic import (like `getAsyncLifecycle(() => import("some_file"))`), Webpack will create a new "chunk" for "some_file", which means that there's a separate JS file that needs to be downloaded. So this series of PRs just get rid of the dynamic imports for most of the "base" components.

In a sense, this is a follow-on for the esm-core V5 update. Prior to V5, we needed to call each module's `setupOpenmrs()` function to determine what extensions, etc., it had, so these dynamic imports helped us defer loading the components until they are actually needed. With V5, the same information is in the routes information, so we already defer loading components until their needed. With that change, reducing the number of Webpack chunks, and thus the number of JS files that need to be downloaded to view a page or extension will speed up the application.

mogoodrich (approved):
Thanks for the explanation @ibacher , LGTM!

src/index.ts
```diff
@@ -1,5 +1,8 @@
-import { getAsyncLifecycle, defineConfigSchema } from "@openmrs/esm-framework";
+import { defineConfigSchema, getSyncLifecycle } from "@openmrs/esm-framework";
 import { configSchema } from "./config-schema";
+import dispensingComponent from "./dispensing.component";
+import dispensingLinkComponent from "./dispensing-link.component";
+import dispeningDashboardComponent from "./dashboard/dispensing-dashboard.component";
```
denniskigen, line 5: Small typo here

ibacher: Ooof... Good catch. I guess I'll do a follow-up.

denniskigen:
@ibacher looks like the implicit rule of thumb is that we’re only changing imports for “base” components. Do you have a rubric for which components ought to be considered, and which ones ought not to? I want to refresh the migration guide in the docs appropriately

ibacher:
> Do you have a rubric for which components ought to be considered, and which ones ought not to?

Basically, I've left these components asynchronous:

1. Modals
2. Workspaces

The guiding rule is something like: "Anything that might need to be rendered on the screen before the user takes an action is synchronous, anything that only renders in response to a user action is asynchronous."

PR #94, opened 2024-01-05
O3-2696 Add ability to configure dispensing app to homepage

CynthiaKamau:
## Requirements

## Summary

This pr will add the dispensing app to the home app. It is a configurable link and implementer's who need the link can add it like using : 

```
"@openmrs/esm-home-app": {
    "extensionSlots": {
      "homepage-dashboard-slot": {
        "add": [
          "dispensing-dashboard-link"
        ]
      }
    }
  }
```

## Screenshots

https://www.loom.com/share/8c2e46878666400390841fa726360af1?sid=e7f8c4b3-b74c-4c0a-9c0c-0938fcd9a5e8

### DIFF
--- a/src/dashboard/dispensing-dashboard-link.component.tsx
+++ b/src/dashboard/dispensing-dashboard-link.component.tsx
@@ -1,0 +1,36 @@
+import { ConfigurableLink } from "@openmrs/esm-framework";
+import React, { useMemo } from "react";
+import classNames from "classnames";
+import { useTranslation } from "react-i18next";
+import { BrowserRouter } from "react-router-dom";
+
+const DispensingDashboardLink = () => {
+  return (
+    <BrowserRouter>
+      <DashboardExtension />
+    </BrowserRouter>
+  );
+};
+
+export default DispensingDashboardLink;
+
+function DashboardExtension() {
+  const { t } = useTranslation();
+  const spaBasePath = `${window.spaBase}/home`;
+  const navLink = useMemo(() => {
+    const pathArray = location.pathname.split("/home");
+    const lastElement = pathArray[pathArray.length - 1];
+    return decodeURIComponent(lastElement);
+  }, [location.pathname]);
+
+  return (
+    <ConfigurableLink
+      className={classNames("cds--side-nav__link", {
+        "active-left-nav-link": navLink.match("dispensing"),
+      })}
+      to={`${spaBasePath}/dispensing`}
+    >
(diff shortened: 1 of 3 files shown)


ibacher:
I would prefer we create the extension but do not make it part of the default homepage. That way implementers that need this functionality can opt-in, which seems simpler, especially as the dispensing app is not a core EMR feature.

CynthiaKamau:
> I would prefer we create the extension but do not make it part of the default homepage. That way implementers that need this functionality can opt-in, which seems simpler, especially as the dispensing app is not a core EMR feature.

Thanks @ibacher , i have made the necessary changes

ibacher (approved):

ebambo (approved):

mseaton:
A bit of a late comment, but I'm curious why this is needed.  There is already a link to the dispensing-link component, which is in the app-menu-slot.  It seems odd that we'd need to add to the dispensing esm a bunch of components that are specific to the slots into which they might be added.  Can't this be driven from configuration in some standard way, or have one component reused?

ibacher:
@mseaton That's a good point. I didn't look too closely. In general the left-menu on the home page should _not_ be a configurable link like this. Configurable links belong in the app-menu-slot. Things on the home page should have a display on the home page rather than redirecting.

ebambo:
@mseaton @ibacher this is how we have designed OHRI from the beginning, you can have a look at [this design](https://app.zeplin.io/project/60d5947dd636aebbd63dce4c/screen/61a5e0a8a332ec10a0f1b40a) of the OHRI Home page. I do agree redirects from the home page are not ideal and if we could embed the dispensing (and the other apps a per the design) would be very helpful for us. Having the apps on the app-menu-slot on the top right is not user friendly as per the comments we received when Paul did the user testing / interviews. Happy to discuss this more in-depth in the appropriate channels.

CynthiaKamau:
> @mseaton That's a good point. I didn't look too closely. In general the left-menu on the home page should _not_ be a configurable link like this. Configurable links belong in the app-menu-slot. Things on the home page should have a display on the home page rather than redirecting.

Thanks @mseaton and @ibacher for the observations. I mostly tried to follow what already exists on the patient management repo for the side nav links and it uses the configuration link eg : https://github.com/openmrs/openmrs-esm-patient-management/blob/41ab54a95887bb914c85ebd88e45a1d5a60213de/packages/esm-appointments-app/src/createDashboardLink.component.tsx#L25 , perhaps later we can work on improving that

PR #107, opened 2024-05-22
(feat) O3-3259 Add ability to deduct stock items while performing dispensing medication

donaldkibet:
## Requirements

## Summary
See https://openmrs.atlassian.net/browse/O3-3259

## Related Issue
For the feature to work, the pharmacist has to be given the privileges required to access stock and inventory rest endpoints.

### DIFF
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -123,9 +123,9 @@ export const configSchema = {
   },
   enableStockDispense: {
     _type: Type.Boolean,
-    _description: 'To enable or disable the deduction of stock during the dispensing process',
-    _default: true,
-  },
+    _description: 'Enable or disable stock deduction during the dispensing process. Requires the stock management module to be installed and configured.',
+    _default: false,
+},
 };
 
 export interface PharmacyConfig {
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -121,6 +121,11 @@ export const configSchema = {
       },
     },
   },
+  enableStockDispense: {
+    _type: Type.Boolean,
+    _description: 'To enable or disable the deduction of stock during the dispensing process',
+    _default: true,
+  },
 };
 
 export interface PharmacyConfig {
@@ -162,4 +167,5 @@ export interface PharmacyConfig {
       uuid: string;
     };
   };
+  enableStockDispense: boolean;
 }
--- a/src/forms/dispense-form.component.tsx
+++ b/src/forms/dispense-form.component.tsx
@@ -28,7 +28,7 @@ import {
 import { updateMedicationRequestFulfillerStatus } from '../medication-request/medication-request.resource';
 import { type PharmacyConfig } from '../config-schema';
 import StockDispense from './stock-dispense/stock-dispense.component';
-import { createStockDispenseRequestPayload, sendStockDispenseRequest } from './stock-dispense/useDispenseStock';
+import { createStockDispenseRequestPayload, sendStockDispenseRequest } from './stock-dispense/stock.resource';
 
 interface DispenseFormProps {
   medicationDispense: MedicationDispense;
@@ -77,30 +77,6 @@ const DispenseForm: React.FC<DispenseFormProps> = ({
               medicationRequestBundle,
               config.dispenseBehavior.restrictTotalQuantityDispensed,
             );
-
-            if (config.enableStockDispense) {
-              const stockDispenseRequestPayload = createStockDispenseRequestPayload(
-                inventoryItem,
-                patientUuid,
-                encounterUuid,
-                medicationDispensePayload,
-              );
-              sendStockDispenseRequest(stockDispenseRequestPayload, abortController).then(
-                () => {
-                  showToast({
-                    title: t('stockDispensed', 'Stock dispensed'),
-                    kind: 'success',
-                    description: t(
-                      'stockDispensedSuccessfully',
-                      'Stock dispensed successfully and batch level updated.',
-                    ),
-                  });
-                },
-                (error) => {
-                  showToast({ title: 'Stock dispense error', kind: 'error', description: error?.message });
-                },
-              );
-            }
             if (getFulfillerStatus(medicationRequestBundle.request) !== newFulfillerStatus) {
               return updateMedicationRequestFulfillerStatus(
                 getUuidFromReference(
@@ -109,6 +85,31 @@ const DispenseForm: React.FC<DispenseFormProps> = ({
                 newFulfillerStatus,
               );
             }
+          }
+          return response;
+        })
+        .then((response) => {
+          const { status } = response;
+          if ((config.enableStockDispense && status === 201) || status === 200) {
+            const stockDispenseRequestPayload = createStockDispenseRequestPayload(
+              inventoryItem,
+              patientUuid,
+              encounterUuid,
+              medicationDispensePayload,
+            );
+            sendStockDispenseRequest(stockDispenseRequestPayload, abortController).then(
+              () => {
+                showToast({
+                  critical: true,
+                  title: t('stockDispensed', 'Stock dispensed'),
+                  kind: 'success',
+                  description: t('stockDispensedSuccessfully', 'Stock dispensed successfully and batch level updated.'),
+                });
+              },
+              (error) => {
+                showToast({ title: 'Stock dispense error', kind: 'error', description: error?.message });
+              },
+            );
           }
           return response;
         })
--- a/src/forms/dispense-form.component.tsx
+++ b/src/forms/dispense-form.component.tsx
@@ -11,7 +11,12 @@ import {
 import { Button, FormLabel, InlineLoading } from '@carbon/react';
 import styles from './forms.scss';
 import { closeOverlay } from '../hooks/useOverlay';
-import { type MedicationDispense, MedicationDispenseStatus, type MedicationRequestBundle } from '../types';
+import {
+  type MedicationDispense,
+  MedicationDispenseStatus,
+  type MedicationRequestBundle,
+  type InventoryItem,
+} from '../types';
 import { saveMedicationDispense } from '../medication-dispense/medication-dispense.resource';
 import MedicationDispenseReview from './medication-dispense-review.component';
 import {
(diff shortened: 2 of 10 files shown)


src/routes.json
```diff
@@ -1,7 +1,8 @@
 {
   "backendDependencies": {
     "fhir2": ">=1.2",
-    "webservices.rest": "^2.2.0"
+    "webservices.rest": "^2.2.0",
+    "stockmanagement": "^1.0.0"
```
jabahum, line 5: change stockmanagement version  to ^1.4.1 >

jabahum (commented):
Please share a gif

src/forms/stock-dispense/stock-dispense.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const toStockDispense = (inventoryItems) => {
+    return t(
+      'stockDispenseDetails',
+      'Batch: {{batchNumber}} - Quantity: {{quantity}} ({{quantityUoM}}) - Expiry: {{expiration}}',
+      {
+        batchNumber: inventoryItems.batchNumber,
+        quantity: Math.floor(inventoryItems.quantity),
+        quantityUoM: inventoryItems.quantityUoM,
+        expiration: formatDate(new Date(inventoryItems.expiration)),
```
jabahum, line 27: format this to MM/YYYY

src/forms/stock-dispense/useDispenseStock.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const url = `/ws/rest/v1/stockmanagement/stockiteminventory?v=default&limit=10&totalCount=true&drugUuid=${drugUuid}&includeBatchNo=true`;
+  const { data, error, isLoading } = useSWR<{ data: { results: Array<InventoryItem> } }>(url, openmrsFetch);
+  return { inventoryItems: data?.data?.results ?? [], error, isLoading };
+};
+
+/**
+ * Interface for the stock dispense request object.
+ */
+export type StockDispenseRequest = {
+  locationUuid: string;
```
jabahum, line 23: this too changes to dispenseLocationUuid

src/forms/stock-dispense/useDispenseStock.tsx
```diff
@@ -0,0 +1,78 @@
...
+  patientUuid: string;
```
jabahum, line 24: this is just patient,order,encounter,stockItem,stockBatch as per new version of stockmanagement 1.4.0

src/forms/stock-dispense/stock-dispense.component.tsx
donaldkibet: Any specific reason why it should marked that way?

src/forms/stock-dispense/useDispenseStock.tsx
donaldkibet: Updated with the requested changes, according to @slubwama [PR](https://github.com/openmrs/openmrs-module-stockmanagement/pull/20) this should be `dispenseLocation `

jabahum: as per the request body and not the response ?

src/config-schema.ts
```diff
@@ -121,6 +121,11 @@ export const configSchema = {
       },
     },
   },
+  enableStockDispense: {
+    _type: Type.Boolean,
+    _description: 'To enable or disable the deduction of stock during the dispensing process',
+    _default: true,
```
mogoodrich, line 127: To maintain the same behavior by default, this default value should be false.

src/routes.json
```diff
@@ -1,7 +1,8 @@
 {
   "backendDependencies": {
     "fhir2": ">=1.2",
-    "webservices.rest": "^2.2.0"
+    "webservices.rest": "^2.2.0",
+    "stockmanagement": "^1.4.1 "
```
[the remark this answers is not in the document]
mogoodrich: Agree with @pirupius , thanks for flagging this.

src/forms/stock-dispense/useDispenseStock.tsx
```diff
@@ -0,0 +1,79 @@
+import useSWR from 'swr';
```
mogoodrich, line 1: I know we've started a pattern of calling something "useDispenseStock" but then including multiple different hooks in it, but I think it would be better to get away from this and call this something like stock.resource.tsx or stock-management.resource.tsx.   Do you agree @denniskigen  @ibacher ?

src/forms/stock-dispense/useDispenseStock.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const useDispenseStock = (drugUuid: string) => {
+  const url = `/ws/rest/v1/stockmanagement/stockiteminventory?v=default&limit=10&totalCount=true&drugUuid=${drugUuid}&includeBatchNo=true`;
+  const { data, error, isLoading } = useSWR<{ data: { results: Array<InventoryItem> } }>(url, openmrsFetch);
+  return { inventoryItems: data?.data?.results ?? [], error, isLoading };
+};
+
+/**
+ * Interface for the stock dispense request object.
+ */
+export type StockDispenseRequest = {
```
mogoodrich, line 22: Is there a reason to include this here instead of in types?

src/forms/stock-dispense/useDispenseStock.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { getUuidFromReference } from '../../utils';
+
+//TODO: Add configuration to retrieve the stock dispense endpoint
+/**
+ * Fetches the inventory items for a given drug UUID.
+ *
+ * @param {string} drugUuid - The UUID of the drug.
+ * @returns {Array} - The inventory items.
+ */
+export const useDispenseStock = (drugUuid: string) => {
```
mogoodrich, line 13: I see you are setting limit=10, will there be a problem if there are more than 10 matching inventory items?

src/forms/dispense-form.component.tsx
```diff
@@ -67,6 +77,30 @@ const DispenseForm: React.FC<DispenseFormProps> = ({
               medicationRequestBundle,
               config.dispenseBehavior.restrictTotalQuantityDispensed,
             );
+
+            if (config.enableStockDispense) {
```
mogoodrich, line 81: This is breaking up the logic around fulfiller status, can we break this up into a separate "then" block, or at least move it up to line 73, before the "const newFulfillerStatus..." line?

Also,   defer to you about the order operations... ie, would you want to move the dispense request *before* the "saveMedicationRequest" so that you *only* update the medication request if the stock update is successful?  Of course, then you run into the opposite risk: that the stock could be updated and then the dispense fails. Would be great to make this a little more transactional, but, admittedly, that is likely difficult.

mogoodrich (changes requested):
Generally looks good @donaldkibet  though I added some requests for changes.

Would be great to see some screen shots of this as well when you get a chance.

src/forms/stock-dispense/useDispenseStock.tsx
ibacher: Indeed, that's substantially more consistent with every other app in the ecosystem.

src/routes.json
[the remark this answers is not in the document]
ibacher: Hmmm... We probably _do_ need a way of indicating optional dependencies, though, right, i.e., if this only works with `stockmanagement` >= 1.4.1, then we do want some way of flagging that for implementations that do use stockmanagement.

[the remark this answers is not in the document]
ibacher: Another thing for the backlog: https://openmrs.atlassian.net/browse/O3-3266

src/forms/stock-dispense/stock-dispense.component.tsx
mogoodrich: Tend to agree with @ibacher

src/config-schema.ts
jabahum: i agree

donaldkibet:
@mogoodrich @ibacher made the requested changes. Attached is an image of the changes

https://github.com/openmrs/openmrs-esm-dispensing-app/assets/28008754/d5fbdbc9-fccc-4068-8a33-bfb9593a844f

src/routes.json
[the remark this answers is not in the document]
donaldkibet: Tried to relax this with 

```json
"stockmanagement": ">=1.4.1"
```

src/forms/stock-dispense/useDispenseStock.tsx
donaldkibet: This was left by mistake, moved to types file

src/routes.json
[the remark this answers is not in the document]
ojwanganto: should it then be removed entirely? If yes, how best can we still have this requirement, at least somewhere in the application?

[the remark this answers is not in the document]
ibacher: Yes. For now, I don't think we can. We need [this feature](https://openmrs.atlassian.net/browse/O3-3266) to properly support that.

donaldkibet:
@mogoodrich removed the stock management requirement from the backend dependencies until this [PR](https://github.com/openmrs/openmrs-esm-core/pull/1015) is merged in. I have also update comments on the config to make it explicit.

src/forms/stock-dispense/useDispenseStock.tsx
jabahum: @slubwama can give us a clear explanation why we need to add this and if its really necessary we find how to pass it from the UI.

mogoodrich (approved):
Thanks @donaldkibet , LGTM!

Feel free to merge.

ojwanganto:
Thanks, @mogoodrich

src/forms/stock-dispense/stock.resource.tsx
```diff
@@ -0,0 +1,67 @@
...
+// For stock dispense to work, stock management module should be installed and configured
+/**
+ * Fetches the inventory items for a given drug UUID.
+ *
+ * @param {string} drugUuid - The UUID of the drug.
+ * @returns {Array} - The inventory items.
+ */
+export const useDispenseStock = (drugUuid: string) => {
+  const session = useSession();
+  const url = `/ws/rest/v1/stockmanagement/stockiteminventory?v=default&totalCount=true&drugUuid=${drugUuid}&includeBatchNo=true&groupBy=LocationStockItemBatchNo&dispenseLocationUuid=${session?.sessionLocation?.uuid}&includeStrength=1&includeConceptRefIds=1&emptyBatch=1&emptyBatchLocationUuid=${session?.sessionLocation?.uuid}&dispenseAtLocation=1`;
```
mogoodrich, line 16: Not to weigh into this after the fact, but, thinking about the TAC call we are on right now, it would be great if, long-term, the stock management module support an FHIR endpoint so that it would hopefully be more standardized.

PR #109, opened 2024-07-16
(fix) O3-3557: Ensure selectable batches have sufficient quantities f…

mogoodrich:
…or dispensing

## Requirements

### DIFF
--- a/src/forms/medication-dispense-review.component.tsx
+++ b/src/forms/medication-dispense-review.component.tsx
@@ -305,11 +305,7 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
           }
           min={0}
           max={config.dispenseBehavior.restrictTotalQuantityDispensed ? quantityRemaining : undefined}
-          value={
-            quantityRemaining <= medicationDispense.quantity.value
-              ? quantityRemaining
-              : medicationDispense.quantity.value
-          }
+          value={medicationDispense.quantity.value}
           onChange={(e) => {
             updateMedicationDispense({
               ...medicationDispense,


src/forms/medication-dispense-review.component.tsx
```diff
@@ -305,11 +305,7 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
           }
           min={0}
           max={config.dispenseBehavior.restrictTotalQuantityDispensed ? quantityRemaining : undefined}
-          value={
-            quantityRemaining <= medicationDispense.quantity.value
-              ? quantityRemaining
-              : medicationDispense.quantity.value
-          }
+          value={medicationDispense.quantity.value}
```
mogoodrich, line 308: So I want to revert this part of the previous commit for O3-3557, as it is restricting the quantity dispensed even if the quantity dispensed flag is not set.  I also think it's confusing, because it's not ever letting the user type in a number greater than the quantity remaining... I think it's better to allow the user to do this but give an error.  Also, as it currently works the "Dispense" button doesn't appear to be enabled until you re-type the number.

I think I do understand the reasoning for this now... prior to this change, when starting a new dispense, the default value on this form is set to whatever the order quantity was... which is likely *not* what you want if the order quantity is less than the quantity remaining.  However, if we want to enforce this, I think we need to do it when the the medication dispense object is initially set up (which I think is in the "initiateMedicationDispenseBody" method in the medication dispense resource).

mogoodrich: Note that admittedly I haven't tested this with the stock management functionality, as I don't have a demo environment for that, not sure if further changes are needed if we revert this to meet your use case.

FelixKiprotich350: So Antony, from your views I think we can keep the filtering of the batches and on selecting that do not change the quantity to dispense ... Let that component remain the way it is.... So generally it will be a filter of batches but let the user input the dispense value

ojwanganto: Thanks, @FelixKiprotich350. Please go ahead and provide your review and approval.

mogoodrich: Yes @FelixKiprotich350  please let me know if this looks okay to you.  

Sorry if I don't understand the how the stock integration functionality it supposed to work.  If the idea is that selecting a batch should change the quantity to dispense, that seems doable, but I would think it should be done via function that is passed into the StockDispense component, like the setAvailableQuantity currently is.  I don't think we can just set the available quantity when the user selected a batch and then determine the dispense amount from that.   Does that make sense?  I'm happy to look at some screen shots to understand the stock dispense functionality better.  fyi @ojwanganto

FelixKiprotich350 (commented):
hello @mogoodrich and @ojwanganto ,  i have checked the code and i have a suggestion, to keep the previous flow of the medication-dispense-review component i can just undo the little changes on that component specifically _the new code that updates the quantity based on the selected batch that i added within this PR_ . In this case we will have achieved the intentions to have only possible selectable and consumable batches and let the user decide on the quantity. incase there is an error let it be raised by that component instead of the batch selection component

src/forms/medication-dispense-review.component.tsx
FelixKiprotich350: yeah sure this looks fine to let the user provide his quantity

FelixKiprotich350 (approved):

ojwanganto (approved):
Thanks, @mogoodrich and @FelixKiprotich350.

mogoodrich:
> hello @mogoodrich and @ojwanganto , i have checked the code and i have a suggestion, to keep the previous flow of the medication-dispense-review component i can just undo the little changes on that component specifically _the new code that updates the quantity based on the selected batch that i added within this PR_ . In this case we will have achieved the intentions to have only possible selectable and consumable batches and let the user decide on the quantity. incase there is an error let it be raised by that component instead of the batch selection component. Just like it worked before the user will be able to edit his/her preferred quantity against the prescribed quantity

Sure that work's for me, @FelixKiprotich350  feel free to issue a new PR for that (or just revert).

src/forms/medication-dispense-review.component.tsx
FelixKiprotich350: hello @mogoodrich i am attaching a short video of how the component works within the stock-dispensing  module
[screen-capture.webm](https://github.com/user-attachments/assets/44fef431-127e-4504-a93c-e50231d82dba)

FelixKiprotich350:
following your suggestions i have undo the changes affecting the dispensequantity and reverted back to the previous state

FelixKiprotich350:
Yes It has a PR have attached a short video on the PR

mogoodrich:
Thanks @FelixKiprotich350  and @ojwanganto , merging this in...  I will take a look at the other PR now...

PR #110, opened 2024-07-18
Fix the auto change of dispensing quantity when a batch is selected

FelixKiprotich350:
## Requirements

## Summary
Fix the auto change of dispensing quantity when a batch is selected

## Screenshots
[screen-capture.webm](https://github.com/user-attachments/assets/4a23ea74-f62e-48c9-abce-c6f480d116db)

## Related Issue
https://openmrs.atlassian.net/browse/O3-3536

### DIFF
--- a/src/forms/stock-dispense/stock-dispense.component.tsx
+++ b/src/forms/stock-dispense/stock-dispense.component.tsx
@@ -9,14 +9,9 @@ type StockDispenseProps = {
   medicationDispense: MedicationDispense;
   updateInventoryItem: (inventoryItem: InventoryItem) => void;
   inventoryItem: InventoryItem;
-  updateAvailableQuantityToMedicationDispenseQuantity: (inventoryItem: InventoryItem) => void;
 };
 
-const StockDispense: React.FC<StockDispenseProps> = ({
-  medicationDispense,
-  updateInventoryItem,
-  updateAvailableQuantityToMedicationDispenseQuantity,
-}) => {
+const StockDispense: React.FC<StockDispenseProps> = ({ medicationDispense, updateInventoryItem }) => {
   const { t } = useTranslation();
   const drugUuid = medicationDispense?.medicationReference?.reference?.split('/')[1];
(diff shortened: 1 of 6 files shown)


FelixKiprotich350:
@mogoodrich  and @ojwanganto have a look at this PR to revert the changes discussed before

mogoodrich:
Oh, sorry, @FelixKiprotich350  I didn't realize that this PR was against my other PR, it looks like it got closed automatically when that other PR got merged in.

Can you reopen if you still want to do this?  I will add one comment on the PR?

src/forms/stock-dispense/stock-dispense.component.tsx
```diff
@@ -110,7 +111,6 @@ const StockDispense: React.FC<StockDispenseProps> = ({
         items={validInventoryItems}
         onChange={({ selectedItem }) => {
           updateInventoryItem(selectedItem);
-          updateAvailableQuantityToMedicationDispenseQuantity(selectedItem);
```
mogoodrich, line 113: To be clear, I'm fine if you want to keep this, and update the amount available to dispense... the problem I wanted to revert is that it was enforcing the amount you could dispense even if you weren't using stock management, and didn't have the restrict quantity enabled.

If you do want to remove this, shouldn't we remove the changes to dispense-form.component.tsx from your initial commit (https://github.com/openmrs/openmrs-esm-dispensing-app/commit/f1b80940fa02927e5072c535aa4217baf7b70cf9) because there is no need to create a setAvailableQuantity function in a "updateAvailalbeQuantityToMedicationDispenseQuantity" function if it no longer being used.  Or am I missing something?

FelixKiprotich350:
Sure I would not wish to create a problem on top of this. Similarly we can
achieve the same by validating the quantity input in the dispense quantity
component.
I think removing this fix is the best because it will leave the code clean
and simple. I will do a follow-up to validate the input later but after
merging this and ensuring everything is working fine as is expected

FelixKiprotich350:
So that the quantity is independent of the context being applied and not
affected by external components like this cone which is coming from
stock-dispense

PR #121, opened 2024-09-27
O3-4043: Dispensing: Allow back dating dispense events for retrospect…

mogoodrich:
…ive entry

## Requirements

## Summary

See: https://openmrs.atlassian.net/browse/O3-4043

## Screenshots

image

## Other

Note that this does not yet cover limiting to providers with certain roles, I need to update the REST Web Services module first to support that.

### DIFF
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -56,10 +56,10 @@ export const configSchema = {
     },
   },
   dispenserProviderRoles: {
-    _type: Type.String,
+    _type: Type.Array,
     _description:
-      'Comma-separated list of provider roles.  If specified, only providers with these roles will be listed in the "Dispensed By" dropdown.  Note that this simply restricts the providers that can be recorded as Dispensers, it does not limit who can create dispense events.',
-    _default: '',
+      'Array of provider roles uuids.  If specified, only providers with these roles will be listed in the "Dispensed By" dropdown.  Note that this simply restricts the providers that can be recorded as Dispensers, it does not limit who can create dispense events.',
+    _default: [],
   },
   medicationRequestExpirationPeriodInDays: {
     _type: Type.Number,
@@ -150,7 +150,7 @@ export interface PharmacyConfig {
     allowModifyingPrescription: boolean;
     restrictTotalQuantityDispensed: boolean;
   };
-  dispenserProviderRoles: string;
+  dispenserProviderRoles: [];
   medicationRequestExpirationPeriodInDays: number;
   locationBehavior: {
     locationColumn: {
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -58,7 +58,7 @@ export const configSchema = {
   dispenserProviderRoles: {
     _type: Type.String,
     _description:
-      'Comma-separated list of provider roles.  If specified, only providers with these roles will be listed in the "Dispensed By" dropdown.  Note that this simply restricts the providers that can be recorded as Dispense, it does not limit who can create dispense events.',
+      'Comma-separated list of provider roles.  If specified, only providers with these roles will be listed in the "Dispensed By" dropdown.  Note that this simply restricts the providers that can be recorded as Dispensers, it does not limit who can create dispense events.',
     _default: '',
   },
   medicationRequestExpirationPeriodInDays: {
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -55,6 +55,12 @@ export const configSchema = {
       _default: false,
     },
   },
+  dispenserProviderRoles: {
+    _type: Type.String,
+    _description:
+      'Comma-separated list of provider roles.  If specified, only providers with these roles will be listed in the "Dispensed By" dropdown.  Note that this simply restricts the providers that can be recorded as Dispense, it does not limit who can create dispense events.',
+    _default: '',
+  },
   medicationRequestExpirationPeriodInDays: {
     _type: Type.Number,
     _description: 'Medication Requests older that this will be considered expired',
@@ -144,6 +150,7 @@ export interface PharmacyConfig {
     allowModifyingPrescription: boolean;
     restrictTotalQuantityDispensed: boolean;
   };
+  dispenserProviderRoles: string;
   medicationRequestExpirationPeriodInDays: number;
   locationBehavior: {
     locationColumn: {
--- a/src/forms/medication-dispense-review.component.test.tsx
+++ b/src/forms/medication-dispense-review.component.test.tsx
@@ -8,6 +8,7 @@ jest.mock('@openmrs/esm-framework', () => {
   return {
     __esModule: true,
     ...originalModule,
+    OpenmrsDatePicker: jest.fn(() => <div />),
     useConfig: jest.fn(() => ({
       dispenseBehavior: {
         allowModifyingPrescription: false,
@@ -129,7 +130,7 @@ describe('Medication Dispense Review Component tests', () => {
     };
 
     const mockUpdate: Function = jest.fn();
-    const { getByText, container } = render(
+    render(
       <MedicationDispenseReview
         medicationDispense={medicationDispense}
         updateMedicationDispense={mockUpdate}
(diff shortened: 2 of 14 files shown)


src/history/history-and-comments.component.tsx
```diff
@@ -214,7 +213,7 @@ const HistoryAndComments: React.FC<{
       {medicationRequestBundles &&
         medicationRequestBundles
           .flatMap((medicationDispenseBundle) => medicationDispenseBundle.dispenses)
-          .sort(sortMedicationDispensesByDateRecorded)
+          .sort(sortMedicationDispensesByWhenHandedOver)
```
mogoodrich, line 216: The FHIR resource for a Medication Dispense supports three dates:
* date recorded (maps to date created in the OpenMRS model)
* when prepared (maps to date prepared in the OpenMRS model)
* when handed over (maps to date handed over in the OpenMRS model)

Previously we were timestamping "when prepared" and "when handed over" to "now" (and date_created was timestamped to "now" on the backend).

Now we allow the user to set the "when prepared" and "when handed over" dates via a single Date of Dispense datepicker (which defaults to now).

I don't remember why we were using the date recorded as the display and sort date, but now that we are allowed orders to be back dated, we need to display and sort by when prepared or when handled over, because these are the more accurate and relevant dates.

We are continuing to not distinguish between "date prepared" and "date handed over" and always set them to the  single "dispense date", but the data module will allow us to expand the Dispensing app to record these two potentially distinct events if we like.

src/forms/medication-dispense-review.component.tsx
```diff
@@ -489,6 +495,48 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
             });
           }}
         />
+
+        <DatePicker
+          datePickerType="single"
+          dateFormat={datePickerFormat}
+          minDate={dayjs(prescriptionDate).startOf('day').toDate()}
```
mseaton, line 502: What happens here if prescriptionDate = null?  That seems like it is possible from the earlier code

mseaton (commented):
One comment, defer to the other react gurus on the rest

src/forms/medication-dispense-review.component.tsx
mogoodrich: Oh, interesting, I hadn't really thought of this, but since dayjs() returns the current datetime, I would think dayjs(null) would return the current datetime as well.

Prescription Date should never be null long-term, but as it is loaded via a hook, it might start null.  In this case, having the min date be today until the prescription is properly loaded like proper behavior?

But I also just might want to *not* display the form until all the data is loaded, I can look into that, as it might fix a weird issue I was seeing as well.

src/forms/medication-dispense-review.component.tsx
```diff
@@ -489,6 +495,48 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
             });
           }}
         />
+
+        <DatePicker
```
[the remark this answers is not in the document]
mogoodrich: Thanks @pirupius  I can look into doing that.

src/forms/medication-dispense-review.component.tsx
chibongho: > I would think dayjs(null) would return the current datetime as well.

I just checked. It actually creates an "Invalid Date" object if the input is null. It should be fine as long as we have a null check prior

src/forms/medication-dispense-review.component.tsx
```diff
@@ -489,6 +500,52 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
             });
           }}
         />
+
+        <OpenmrsDatePicker
```
mogoodrich, line 504: @denniskigen  @ibacher  @pirupius  would be great to get your thoughts on whether I'm using this date component correctly, specifically when setting and updating the date.

One weird thing I'm seeing is this display in the datepicker (12:00AM), which I'm guessing happens because I'm passing a full date (with time) in?

image

And also, do we have control of the date display format, or is this controller by the locale?  We try to avoid the American Month-Day-Year, but maybe it is just displaying that because I'm in the US?

image

mogoodrich:
Make some changes to address code review comment, and also now have support for limiting by dispensers by provider role (pending merge of https://openmrs.atlassian.net/browse/PROV-108)

Ready for re-review

src/config-schema.ts
```diff
@@ -55,6 +55,12 @@ export const configSchema = {
       _default: false,
     },
   },
+  dispenserProviderRoles: {
+    _type: Type.String,
```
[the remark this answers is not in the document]
mogoodrich: Fair enough, even if we go with the comma-separated list on the REST web services side to match other conventions, we should use an array here.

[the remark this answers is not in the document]
mogoodrich: Fixed!

src/forms/medication-dispense-review.component.tsx
mseaton: Let's start a separate ticket / conversation about this @mogoodrich

chibongho (commented):
We should resolve the OpnemrsDatePicker, but looks good otherwise.

src/forms/medication-dispense-review.component.tsx
pirupius: Yes @mogoodrich the format is based on locale

src/forms/medication-dispense-review.component.tsx
```diff
@@ -489,6 +500,52 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
...
+          onChange={(selectedDate) => {
+            const currentDate = dayjs(medicationDispense.whenHandedOver);
+            updateMedicationDispense({
+              ...medicationDispense,
+              whenHandedOver: isSameDay(currentDate, selectedDate)
+                ? currentDate.toISOString()
+                : selectedDate.toString(), // to preserve any time component, only update if the day actually changes
+            });
+          }}
+          value={dayjs(medicationDispense.whenHandedOver).startOf('day').toDate()}></OpenmrsDatePicker>
```
pirupius, line 518: The time (12:00AM) being displayed might be because you added `startOf('day')` here

src/forms/medication-dispense-review.component.test.tsx
```diff
@@ -2,23 +2,24 @@ import React from 'react';
...
+      allowModifyingPrescription: false,
+      restrictTotalQuantityDispensed: false,
+    },
+    valueSets: {
+      substitutionType: { uuid: '123' },
+      substitutionReason: { uuid: 'abc' },
+    },
+  });
+
+  mockOpenmrsDatePicker.mockReturnValue(<div />);
```
mogoodrich, line 22: Updated, fyi @chibongho , this is much cleaner!

src/forms/medication-dispense-review.component.tsx
mogoodrich: Thanks @pirupius s 

@pirupius @mseaton  I couldn't reproduce this behavior ("Oct 12:00AM") today, I'm wondering if it had anything to do with the fact that it was the first of the month when I was testing?

I did remove the "startOf('day')" as you suggested @pirupius ... I believe that was only something I added in hopes it might fix the above issue.

But I also agree with @mseaton , we should likely start another ticket/convo about the OpenmrsDataPicker and not have it block this ticket.

mogoodrich:
@chibongho  I fixed the mock thing, thanks!  I also made one tweak to the OpenmrsDatePickup, but do agree with @mseaton  that other issues with the datepicker should probably be ticketed as a separate ticket.  Do you think I've gotten far enough to get an approval on these changes and get them merged in?

chibongho (approved):

chibongho:
> @chibongho I fixed the mock thing, thanks! I also made one tweak to the OpenmrsDatePickup, but do agree with @mseaton that other issues with the datepicker should probably be ticketed as a separate ticket. Do you think I've gotten far enough to get an approval on these changes and get them merged in?

Yeah, fine with me. Thanks!

pirupius (approved):
Sounds good me @mogoodrich

mogoodrich (commented):
Merging, thanks all!

PR #123, opened 2024-11-18
(feat) O3-4206: Add ability to view patient diagnoses and active conditions

Omoshlawi:
## Requirements

## Summary

As discused in [slack](https://openmrs.slack.com/archives/C02P790SM7X/p1723007479471759) and based 

- Added Configurable  patient active conditions (Disabled by default)
- Made patient Visit Diagnoses configurable (disabled by default)
- Added extra patient information tab in vertical tab menu whose content is a slot allowing different implementors to plug in any extra information required
- The tab is also configurable and disabled by default

## Screenshots

https://github.com/user-attachments/assets/18640401-7b59-44ee-ab7c-fc9abc508993

## Related Issue
https://openmrs.atlassian.net/browse/O3-4206


src/diagnoses/diagnoses.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+      </strong>
+      <br />
+      {diagnoses.length ? (
+        <>
+          {diagnoses.map(({ text }, index) => (
+            <Tag key={index}>{text}</Tag>
+          ))}
+        </>
+      ) : (
+        <>{t('noDiagnoses', 'No diagnoses found')}</>
```
mogoodrich, line 39: I would go with just displaying nothing at all if no diagnoses are found?  This is would make it more somewhat more compatible with implementations that are not yet storing diagnoses using the "encounter diagnosis" module?

mogoodrich (commented):
Thanks so much @Omoshlawi .... will generally defer to others on this, with some thoughts:

* Looks like this only displays the diagnoses records in the visit, which likely makes sense, but would be good for others to weigh in 
* I don't love the way the diagnoses are displayed in the screen capture, but will defer to the UX team if we are following display standards correctly
* In my opinion, we should hide the diagnoses display if no diagnoses are found (instead of displaying "no diagnoses found")

I will add some others to the reviewers.

mogoodrich:
Are Paul and/or Ciaran on Github and can they be flagged here to review the UI?  Otherwise, @Omoshlawi  can you add the screen capture you collected to the linked ticket and flag them there for review?  thanks!

denniskigen:
@mogoodrich Ping @paulsonder @ciaranduffy

Omoshlawi:
> Are Paul and/or Ciaran on Github and can they be flagged here to review the UI? Otherwise, @Omoshlawi can you add the screen capture you collected to the linked ticket and flag them there for review? thanks!

Okay, sure

mogoodrich (commented):
Thanks so much, this is looking really good!

So I added one question/comment. Beyond that:

* eventually we want to be able to pull in the conditions and diagnoses via a standard resource provided by core or the patient chart instead of having to recreate a resource in each app that we want to display conditions/diagnoses in, but I don't want to be a blocker in getting this merged in if we aren't there yet.  Thoughts @ibacher  @denniskigen ?

* Also, overall, it would be good for someone else to get eyes on this with more O3 experience that me

src/components/empty-state.component.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Layer } from '@carbon/react';
+import { Report } from '@carbon/react/icons';
+import styles from './empty-state.scss';
+
+type EmptyState = {
```
ibacher, line 6: We don't need a whole component for this. Empty states should match the design of the thing that they are the empty state for. Generic empty states are only useful in contexts where we have many similar widgets.

src/components/empty-state.scss
```diff
@@ -0,0 +1,50 @@
...
+  background-color: colors.$white;
+  margin: layout.$spacing-03 0 0 0;
+}
+
+.title {
+  @include type.type-style('heading-01');
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  row-gap: 1.5rem;
```
ibacher, line 16: Please use Carbon spacing tokens for spacing in SCSS `row-gap`, `width`, etc.

src/components/empty-state.scss
```diff
@@ -0,0 +1,50 @@
...
+  position: relative;
+  color: colors.$gray-50;
+
+  &::after {
+    content: '';
+    display: block;
+    width: 2rem;
+    border-bottom: 0.375rem solid var(--brand-03);
+    position: absolute;
+    bottom: -0.75rem;
```
ibacher, line 26: This looks like a hack. Why do we need this here?

src/conditions/conditions.resource.ts
```diff
@@ -0,0 +1,133 @@
...
+  display: string;
+  patient: string;
+  onsetDateTime: string;
+  recordedDate: string;
+  recorder: string;
+}
+
+export const usePatientConditions = (patientUuid: string) => {
+  const { showPatientConditions } = useConfig<PharmacyConfig>();
+  const url = `${fhirBaseUrl}/Condition?patient=${patientUuid}&_count=100&_summary=data`;
```
ibacher, line 104: What happens if there are more than 100? Shouldn't we have some sort of paging here?

src/config-schema.ts
```diff
@@ -133,6 +133,21 @@ export const configSchema = {
...
+    _type: Type.Boolean,
+    _description: 'Enable or disable viewing of patient visit diagnoses',
+    _default: false,
+  },
+  showExtraPatientInformationSlot: {
+    _type: Type.Boolean,
+    _description: 'Enable or disable viewing of patient extra information slot diagnoses',
+    _default: false,
+  },
+  showPatientConditions: {
```
ibacher, line 146: We also don't need _configuration_ options where things are exposed as extensions. Please remove these.

src/diagnoses/diagnoses.component.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <StructuredListRow head>
+            <StructuredListCell head>
+              {t('diagnoses', 'Diagnoses')} {`(${diagnoses.length})`}
+            </StructuredListCell>
+            <StructuredListCell head>{t('status', 'Status')}</StructuredListCell>
+          </StructuredListRow>
+        </StructuredListHead>
+        <StructuredListBody>
+          {diagnoses.map(({ certainty, text }) => (
+            <StructuredListRow>
```
ibacher, line 61: Any time you call `map()` in a React component, the component _must_ have a key.

src/diagnoses/diagnoses.component.tsx
```diff
@@ -0,0 +1,72 @@
...
+      />
+    );
+
+  return (
+    <Layer className={styles.diagnosesContainer}>
+      <StructuredListWrapper>
+        <StructuredListHead>
+          <StructuredListRow head>
+            <StructuredListCell head>
+              {t('diagnoses', 'Diagnoses')} {`(${diagnoses.length})`}
```
ibacher, line 54: Is `diagnoses.length` something that anyone asked for here? Why aren't we just using a data table or something?

src/routes.json
```diff
@@ -19,6 +19,20 @@
       "online": true,
       "offline": true
     },
+    {
+      "name": "patient-diagnoses",
+      "slot": "extra-patient-information-slot",
```
ibacher, line 24: These extensions should _not_ be bound to any slot by default. That way, they are already able to be enabled or disabled by configuration.

src/routes.json
```diff
@@ -19,6 +19,20 @@
...
+      "component": "patientDiagnoses",
+      "online": true,
+      "offline": true
```
ibacher, line 27: Also, I'd probably just leave `offline` alone.

ibacher (changes requested):
Some more technical comments that what I've put on Slack.

src/config-schema.ts
```diff
@@ -138,6 +138,16 @@ export const configSchema = {
     _description: 'Enable or disable viewing of patient visit diagnoses',
     _default: false,
   },
+  showExtraPatientInformationSlot: {
+    _type: Type.Boolean,
+    _description: 'Enable or disable viewing of patient extra information slot diagnoses',
+    _default: false,
```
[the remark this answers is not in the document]
Omoshlawi: Ok, Removed

src/components/empty-state.component.tsx
Omoshlawi: i didn't quite understand this, kindly explain further, do you mean something like this for consistent desgin in despensing app
Screenshot 2025-01-10 at 12 14 21

src/config-schema.ts
Omoshlawi: Yes, this makes much sense, did just that, the extensions will be added from the config-app, for now to demo just added them from implementor tool

https://github.com/user-attachments/assets/2c9b2d62-507d-43bc-b41d-54254362b3a6

src/diagnoses/diagnoses.component.tsx
Omoshlawi: Removed and replaced by data tables, bellow is a screenshot of how it looks now
Screenshot 2025-01-10 at 19 15 18

ojwanganto: Thanks, @Omoshlawi for the good work. I do think that for pharmacists, a more compressed presentation is all they need. I would suggest that you revert to the initial implementation that looked like the below:

image
@ibacher and I are aligned on the same

donaldkibet:
ping @mogoodrich @ibacher

src/diagnoses/diagnoses.component.tsx
ojwanganto: @Omoshlawi could you kindly provide a screenshot that reverts this change?

src/conditions/conditions.resource.ts
ibacher: So, `_offset` is a non-standard FHIR parameter and should not be sent in the initial request (as it's not in the actual code). With that parameter, the `next` and `previous` links are weird, without it they should work.

ibacher (changes requested):
So, as I discussed with @ojwanganto and was mentioned here, I don't think that displaying conditions and diagnoses in data tables is an effective use of screen space for pharmacists. The previous view, with tags, is much more compact and don't see a compelling reason to render tables here.

Can we please revert to using tags for diagnoses and do something similar for conditions?

ibacher:
Key thing here: as a pharmacist, I need to verify that the ordered drug is a valid treatment for the patient. To do that, I need to understand what we are treating the patient for, so it's helpful to be able to see the patient's diagnoses and conditions that relate or may relate to the ordered drug. However, to do this, I just basically need a quick overview of the _names_ of any (confirmed) diagnosed conditions or problems. The table doesn't generally include any useful details and those could easily be moved to a tooltip or something similar, allowing us to better compress the data we're feeding to users so they can see it all on one screen.

src/diagnoses/diagnoses.component.tsx
Omoshlawi: Here is.I have reserved the configurability through implementor config-app

Screenshot 2025-01-26 at 11 00 15

Omoshlawi:
> Key thing here: as a pharmacist, I need to verify that the ordered drug is a valid treatment for the patient. To do that, I need to understand what we are treating the patient for, so it's helpful to be able to see the patient's diagnoses and conditions that relate or may relate to the ordered drug. However, to do this, I just basically need a quick overview of the _names_ of any (confirmed) diagnosed conditions or problems. The table doesn't generally include any useful details and those could easily be moved to a tooltip or something similar, allowing us to better compress the data we're feeding to users so they can see it all on one screen.



> Key thing here: as a pharmacist, I need to verify that the ordered drug is a valid treatment for the patient. To do that, I need to understand what we are treating the patient for, so it's helpful to be able to see the patient's diagnoses and conditions that relate or may relate to the ordered drug. However, to do this, I just basically need a quick overview of the _names_ of any (confirmed) diagnosed conditions or problems. The table doesn't generally include any useful details and those could easily be moved to a tooltip or something similar, allowing us to better compress the data we're feeding to users so they can see it all on one screen.

Screenshot 2025-01-26 at 11 00 15

Omoshlawi:
> So, _offset is a non-standard FHIR parameter and should not be sent in the initial request (as it's not in the actual code). With that parameter, the next and previous links are weird, without it they should work.


I have removed it and here is how the `next` looks like, still weird.And also now that we are using tags, hows pagination handled??
Screenshot 2025-01-26 at 11 19 30

src/diagnoses/diagnoses.component.tsx
ojwanganto: Please add screenshots which are a bit visible

Omoshlawi: Screenshot 2025-01-27 at 18 45 01

ibacher:
> I have removed it and here is how the next looks like, still weird

Weird, but correct. URLs are defined for `next` and `previous` are opaque (in the background, HAPI retains the search query in memory and returns a locator to get the next page from that in-memory instance).

>  hows pagination handled??

Precisely by following the `next` or `previous` links. You can also use the next link but substitute the number of the page for the `_getpagesoffset=<n>` parameter, but I can't guarantee that will always work. I'm pretty sure Chi-Bong wrote some helpers in core for paginated FHIR data that handle things correctly. 

By the way, if you send `_count=1`, you are asking for each page to be a single data item, which seems wasteful.

makombe:
Thanks @Omoshlawi . @ibacher Please have another look at this if it is now ready to go, thanks

paulsonder:
@makombe, thanks for sharing.

I have a few comments regarding the design in the video up top...

1. Could we change the label from 'Extra patient Info' to 'Conditions and diagnoses'?
2. It looks like the two tables are custom components, rather than reusing existing Carbon Tables or Structured lists? Could we try and use Structured lists here?

ibacher:
@makombe Can we also drop the "warning" style text for when there are no conditions or diagnoses and maybe adopt the pill styling that we have for allergies. Ideally, the conditions and diagnoses should stand out more than text noting the absence of information.

Omoshlawi:
> @makombe Can we also drop the "warning" style text for when there are no conditions or diagnoses and maybe adopt the pill styling that we have for allergies. Ideally, the conditions and diagnoses should stand out more than text noting the absence of information.

thanks @ibacher @paulsonder @makombe , here is the latest

Screenshot 2025-03-13 at 13 12 39

Omoshlawi:
> @makombe, thanks for sharing.
> 
> I have a few comments regarding the design in the video up top...
> 
> 1. Could we change the label from 'Extra patient Info' to 'Conditions and diagnoses'?
> 2. It looks like the two tables are custom components, rather than reusing existing Carbon Tables or Structured lists? Could we try and use Structured lists here?

Thanks @paulsonder , for suggestion (2), we (@ojwanganto @ibacher ) resolved to using ui component that covers little space hence did away with structured list and data table

paulsonder:
Hi @Omoshlawi - thanks for the update!

If there are no diagnoses, why not show an empty state or a data table / structured list?

I don't think this is the correct application of an Alert, which should call attention to something. In this case, we're giving the impression that no data is the same as some data, like an Allergy, which _we do_ want to visually emphasise. 

Space is not at a premium here, we should try to prioritise recognisable, consistent patterns for displaying information and tables and lists are used heavily in O3 and don't require custom code to be maintained.

CC @ciaranduffy

Omoshlawi:
@paulsonder , to make sure we are in one page, you suggest we have something as in which options

1.Data table see 
- [With data](https://github.com/openmrs/openmrs-esm-dispensing-app/pull/123#discussion_r1910603924)
- [Empty state](https://github.com/openmrs/openmrs-esm-dispensing-app/pull/123#discussion_r1910595320)

2. Structured list 

https://github.com/user-attachments/assets/18640401-7b59-44ee-ab7c-fc9abc508993

paulsonder:
@Omoshlawi Thanks for sharing the video... looking good... I think we should try to re-use the **Data Tile** component. I can see here the **Empty State** does, but then switches to something more custom when there is data.

You can get full design guidance on those tiles here: https://zeroheight.com/23a080e38/p/41d6df-data-tile

src/diagnoses/diagnoses.resource.ts
```diff
@@ -0,0 +1,30 @@
+import { type FetchResponse, openmrsFetch, restBaseUrl, type Visit } from '@openmrs/esm-framework';
+import { useMemo } from 'react';
+import useSWR from 'swr';
+
+export const usePatientDiagnosis = (encounterUuid: string) => {
+  const customRepresentation =
+    'custom:(uuid,display,visit:(uuid,patient,encounters:(uuid,diagnoses:(uuid,display,certainty,diagnosis:(coded:(uuid,display))),encounterDatetime,encounterType:(uuid,display),encounterProviders:(uuid,display,provider:(uuid,person:(uuid,display)))),location:(uuid,name,display),visitType:(uuid,name,display),startDatetime,stopDatetime))';
```
mogoodrich, line 7: Are we using all these elements? It looks like we are just using visit?.encounter?.diagnoses?

Omoshlawi: True, thanks @mogoodrich , fixed

src/conditions/conditions.component.tsx
```diff
@@ -0,0 +1,95 @@
...
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+  Pagination,
+} from '@carbon/react';
+import { ErrorState, formatDate, parseDate, usePagination } from '@openmrs/esm-framework';
+import { CardHeader, EmptyState } from '@openmrs/esm-patient-common-lib';
```
ibacher, line 14: Seems preferable that we move these from common-lib to the styleguide. Adding a dependency on common-lib here is ideally not something we would do.

src/conditions/conditions.component.tsx
```diff
@@ -0,0 +1,95 @@
...
+type PatientConditionsProps = {
+  patientUuid: string;
+  encounterUuid: string;
+};
+
+const PatientConditions: React.FC<PatientConditionsProps> = ({ encounterUuid, patientUuid }) => {
+  const { t } = useTranslation();
+  const { conditions, error, isLoading, mutate } = usePatientConditions(patientUuid);
+  const [pageSize, setPageSize] = useState(3);
+  const pageSizesOptions = useMemo(() => [3, 5, 10, 20, 50, 100], []);
```
ibacher, line 29: `useMemo()` seems like overkill here? Why not just move `pageSizeOptions` outside of the component.

src/diagnoses/diagnoses.scss
```diff
@@ -0,0 +1,10 @@
+@use '@carbon/layout';
+@use '@carbon/colors';
+@use '@carbon/type';
+@import '~@openmrs/esm-styleguide/src/vars';
```
ibacher, line 4:
```suggestion
@use '@openmrs/esm-styleguide/src/vars' as *;
```

src/prescriptions/prescription-expanded.component.tsx
```diff
@@ -23,18 +23,14 @@ const PrescriptionExpanded: React.FC<{
       name: t('prescriptionDetails', 'Prescription details'),
       component: <PrescriptionDetails encounterUuid={encounterUuid} patientUuid={patientUuid} />,
     },
+    {
+      name: t('conditionsAndDiagnoses', 'Conditions and diagnoses'),
+      component: <ExtensionSlot name="extra-patient-information-slot" state={{ patientUuid, encounterUuid }} />,
```
ibacher, line 28:
```suggestion
      component: <ExtensionSlot name="dispensing-condition-and-diagnoses" state={{ patientUuid, encounterUuid }} />,
```

src/index.ts
```diff
@@ -44,3 +46,6 @@ export const dispenseWorkspace = getAsyncLifecycle(() => import('./forms/dispens
 export const pauseDispenseWorkspace = getAsyncLifecycle(() => import('./forms/pause-dispense-form.workspace'), options);
 
 export const printPrescriptionPreviewModal = getSyncLifecycle(PrescriptionPrintPreviewModal, options);
+
+export const patientDiagnoses = getSyncLifecycle(PatientDiagnoses, options);
+export const patientConditions = getSyncLifecycle(PatientConditions, options);
```
ibacher, line 51: Please use `getAsyncLifecycle()` here.

src/conditions/conditions.scss
```diff
@@ -0,0 +1,10 @@
+@use '@carbon/layout';
+@use '@carbon/colors';
+@use '@carbon/type';
+@import '~@openmrs/esm-styleguide/src/vars';
```
ibacher, line 4:
```suggestion
@use '@openmrs/esm-styleguide/src/vars' as *;
```

src/conditions/conditions.resource.ts
```diff
@@ -0,0 +1,97 @@
...
+  recordedDate: string;
+  recorder: string;
+}
+
+export const usePatientConditions = (patientUuid: string) => {
+  const url = `${fhirBaseUrl}/Condition?patient=${patientUuid}&_summary=data`;
+  const { data, isLoading, mutate, error } = useFhirFetchAll<Resource>(url);
+
+  const conditions = useMemo(() => {
+    return data?.reduce<Array<Condition>>((prev, entry) => {
```
ibacher, line 75: Wouldn't it be better to just declare this as the type for `useFhirFetchAll()` and not need the case here?

src/conditions/conditions.resource.ts
```diff
@@ -0,0 +1,97 @@
...
+      if (entry?.resourceType === 'Condition') {
+        const condition: Condition = {
+          id: entry.id,
+          display: entry?.code?.text,
+          onsetDateTime: entry?.onsetDateTime,
+          patient: entry?.subject?.display,
+          recordedDate: entry?.recordedDate,
+          recorder: entry?.recorder?.display,
+          status: entry?.clinicalStatus?.coding[0]?.code as any,
```
ibacher, line 84: Why the cast to `any` here?

src/conditions/conditions.resource.ts
```diff
@@ -0,0 +1,97 @@
...
+        };
+        return [...prev, condition];
+      }
+      return prev;
+    }, []);
+  }, [data]);
+  return {
```
ibacher, line 91:
```suggestion
  }, [data]);
  
  return {
```

src/conditions/conditions.resource.ts
```diff
@@ -0,0 +1,97 @@
...
+        const condition: Condition = {
+          id: entry.id,
+          display: entry?.code?.text,
+          onsetDateTime: entry?.onsetDateTime,
+          patient: entry?.subject?.display,
+          recordedDate: entry?.recordedDate,
+          recorder: entry?.recorder?.display,
+          status: entry?.clinicalStatus?.coding[0]?.code as any,
+        };
+        return [...prev, condition];
```
ibacher, line 86:
```suggestion
        };
        
        return [...prev, condition];
```

src/conditions/conditions.resource.ts
```diff
@@ -0,0 +1,97 @@
...
+      }
+      return prev;
```
ibacher, line 88:
```suggestion
      }
      
      return prev;
```

src/conditions/conditions.resource.ts
ibacher: Why do we do a splat and copy? It seems like `prev.push(condition)` would work just as well and not involve the need to copy things.

src/conditions/conditions.resource.ts
```diff
@@ -0,0 +1,97 @@
...
+    }, []);
+  }, [data]);
+  return {
+    conditions: (conditions ?? []).filter((condition) => condition.status === 'active'),
```
ibacher, line 92: Why isn't this part of the query we're sending to the backend? Much better to filter on the backend than here.

src/conditions/conditions.component.tsx
```diff
@@ -0,0 +1,95 @@
...
+    return [
+      { header: t('conditions', 'Condition'), key: 'display' },
+      { header: t('onsetDate', 'Onset Date'), key: 'onsetDateTime' },
+    ];
+  }, [t]);
+
+  const tablerows = useMemo(() => {
+    return (results ?? []).map((c) => ({ ...c, onsetDateTime: formatDate(parseDate(c.onsetDateTime)) }));
+  }, [results]);
+  const title = t('activecondition', 'Active Condition');
```
ibacher, line 41:
```suggestion
  }, [results]);
  
  const title = t('activeCondition', 'Active Condition');
```

src/conditions/conditions.component.tsx
```diff
@@ -0,0 +1,95 @@
...
+            <TableHead>
+              <TableRow>
+                {headers.map((header) => (
+                  <TableHeader {...getHeaderProps({ header })}>{header.header}</TableHeader>
+                ))}
+              </TableRow>
+            </TableHead>
+            <TableBody>
+              {rows.map((row) => (
+                <TableRow {...getRowProps({ row })}>
```
ibacher, line 71:
```suggestion
              {rows.map((row, i) => (
                <TableRow {...getRowProps({ row })} key={i}>
```

src/conditions/conditions.component.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <CardHeader title={title}>
+        <React.Fragment />
+      </CardHeader>
+      <DataTable useZebraStyles rows={tablerows} headers={headers}>
+        {({ rows, headers, getTableProps, getHeaderProps, getRowProps }) => (
+          <Table {...getTableProps()}>
+            <TableHead>
+              <TableRow>
+                {headers.map((header) => (
+                  <TableHeader {...getHeaderProps({ header })}>{header.header}</TableHeader>
```
ibacher, line 65:
```suggestion
                {headers.map((header, i) => (
                  <TableHeader {...getHeaderProps({ header })}, key={i}>{header.header}</TableHeader>
```

src/conditions/conditions.component.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const { results, totalPages, currentPage, goTo } = usePagination(conditions, pageSize);
+  const headers = useMemo(() => {
+    return [
+      { header: t('conditions', 'Condition'), key: 'display' },
+      { header: t('onsetDate', 'Onset Date'), key: 'onsetDateTime' },
+    ];
+  }, [t]);
+
+  const tablerows = useMemo(() => {
+    return (results ?? []).map((c) => ({ ...c, onsetDateTime: formatDate(parseDate(c.onsetDateTime)) }));
```
ibacher, line 39: Why not do this in the `usePatientConditions()` hook?

src/conditions/conditions.component.tsx
ibacher: We should ticket this for later.

ibacher (changes requested):
Thanks @Omoshlawi! A few technical fixes and this should be good to go.

src/conditions/conditions.resource.ts
Omoshlawi: The `useFhirFetchAll` return fhir resource which i then map to `Condition` which is a simple object with only few fields of interest

src/routes.json
```diff
@@ -20,6 +20,18 @@
       "online": true,
       "offline": true
     },
+    {
+      "name": "patient-diagnoses",
+      "component": "patientDiagnoses",
```
ibacher, line 25: Shouldn't these by default be added to the slot we added for them?

ibacher (approved):

PR #126, opened 2024-11-27
(feat) O3-4222: Display patient age in prescriptions table

Omoshlawi:
## Requirements

## Screenshots

https://github.com/user-attachments/assets/fdf61d44-830e-4a0c-9d37-baec055aff01

## Related Issue
https://openmrs.atlassian.net/browse/O3-4222


mogoodrich:
Although adding the patient age is value, I would rather see it added to the "Patient name" column (and perhaps rename that column "Patient") instead of adding a new "Patient age" column.  Thoughts @fiona855  @denniskigen  @ibacher @paulsonder ?

makombe:
> Although adding the patient age is value, I would rather see it added to the "Patient name" column (and perhaps rename that column "Patient") instead of adding a new "Patient age" column. Thoughts @fiona855 @denniskigen @ibacher @paulsonder ?

@Correct that was the agreement. @Omoshlawi please check slack conversation on dispensing channel

ibacher:
As a reminder, Slack is not an official record of decisions made. If agreements like this are made on Slack, we need to update the corresponding Jira ticket with the new requirements.

ojwanganto:
> Although adding the patient age is value, I would rather see it added to the "Patient name" column (and perhaps rename that column "Patient") instead of adding a new "Patient age" column. Thoughts @fiona855 @denniskigen @ibacher @paulsonder ?

+1

paulsonder:
> Although adding the patient age is value, I would rather see it added to the "Patient name" column (and perhaps rename that column "Patient") instead of adding a new "Patient age" column. Thoughts @fiona855 @denniskigen @ibacher @paulsonder ?

+ 1

Omoshlawi:
thanks @makombe @mogoodrich @ojwanganto nd @paulsonder .

concatenated patient age to patient name column display
Screenshot 2024-12-02 at 08 05 15

src/patient/patient.resources.ts
```diff
@@ -0,0 +1,19 @@
...
+    age: data?.data?.age,
+    error,
+    isLoading,
+    mutate,
+  };
+};
+
+export function concatAgePatientDisplay(input: string, age: number): string | null {
+  const attrIndex = input.lastIndexOf(')');
+  return input.slice(0, attrIndex) + ', Age: ' + age + input.slice(attrIndex);
```
mogoodrich, line 18: Minor, but please use "t()" here to localize "Age"

mogoodrich: Also, as this is a  function around formatting and display, let's put this in the patient-info-cell component, not the resource.

src/patient/patient-info-cell.component.tsx
```diff
@@ -0,0 +1,18 @@
...
+    name: string;
+    uuid: string;
+  };
+};
+
+const PatientInfoCell: React.FC<PatientInfoCellProps> = ({ patient: { name, uuid } }) => {
+  const { error, isLoading, age } = usePatientAge(uuid);
+  if (isLoading || error) return <>{name}</>;
+  const patient = concatAgePatientDisplay(name, age);
+  return <>{patient}</>;
```
mogoodrich, line 15: Shouldn't this be {name}?

src/patient/patient.resources.ts
mogoodrich: Sorry, one other thing... but if the lastIndexOf(')') returns -1 for some reason, let's not do the slice and simply append "Age: " + age to the end?  (Just in case the display format changes at some point.

mogoodrich (changes requested):
Thanks!  Generally looks great!  I added a few comments, but hopefully they are all minor.

src/patient/patient.resources.ts
Omoshlawi: Thanks @mogoodrich , incorporated your suggestions

src/patient/patient-info-cell.component.tsx
Omoshlawi: it was ok, but i have renamed the variables appropriately to prevent confusion.renamed the {patient} to {displayWithAge} and {name} to {display} which is the patient name before age is concatenated

mogoodrich (approved):

PR #140, opened 2024-12-17
(fix) Update default valueSet UUIDs

denniskigen:
## Requirements

## Summary

Updates default UUIDs for valueSets in the config schema to match UUIDs that exist in the dev3 environment by default.

### DIFF
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -1,25 +1,5 @@
 import { Type } from '@openmrs/esm-framework';
 
-/**
- * This is the config schema. It expects a configuration object which
- * looks like this:
- *
- * ```json
- * { "casualGreeting": true, "whoToGreet": ["Mom"] }
- * ```
- *
- * In OpenMRS Microfrontends, all config parameters are optional. Thus,
- * all elements must have a reasonable default. A good default is one
- * that works well with the reference application.
- *
- * To understand the schema below, please read the configuration system
- * documentation:
- *   https://openmrs.github.io/openmrs-esm-core/#/main/config
- * Note especially the section "How do I make my module configurable?"
- *   https://openmrs.github.io/openmrs-esm-core/#/main/config?id=im-developing-an-esm-module-how-do-i-make-it-configurable
- * and the Schema Reference
- *   https://openmrs.github.io/openmrs-esm-core/#/main/config?id=schema-reference
- */
 export const configSchema = {
   appName: {
     _type: Type.String,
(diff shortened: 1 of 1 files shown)


denniskigen:
@mogoodrich this is a breaking change for you. I assume that the default UUIDs that are currently set exist in the PIH EMR, so this change would require you to set those values via config instead for your distro.

denniskigen:
Is it intentional that `reasonForPause` and `reasonForClose`  valueSets use the same CIEL concept in the dispensing app config schema? See the descriptions for the respective properties in the config schema [here](https://github.com/openmrs/openmrs-esm-dispensing-app/blob/8bacf9aa86af909b145585fef1ee848519febd5f/src/config-schema.ts#L100) and [here](https://github.com/openmrs/openmrs-esm-dispensing-app/blob/8bacf9aa86af909b145585fef1ee848519febd5f/src/config-schema.ts#L109).

chibongho:
Deferring to @mogoodrich , but I think this change won't break us as we explicit set the config values here:

https://github.com/PIH/openmrs-config-pihemr/blob/d02efb1c445406900cb5e87a1bc6034a69a71868/configuration/frontend/base-config.json#L47

mseaton (commented):
I generally disapprove of any and all default values that refer to specific metadata, and have long felt that the refapp needs to start explicitly configuring ESMs in it's config, and not expecting all ESMs to work in it by default.  I want to break this existing (anti)pattern.  That said, if we are going to insist on defaults, they should be what Mark added I think - which are the default values provided by CIEL - not by some arbitrarily different values in the refapp.

ibacher:
@mseaton The goal here is actually switching from what seem to be PIH-specific UUIDs to CIEL-provided UUIDs. We're not inventing new metadata for the RefApp.

mseaton:
> The goal here is actually switching from what seem to be PIH-specific UUIDs to CIEL-provided UUIDs. We're not inventing new metadata for the RefApp.

If they are the CIEL UUIDs that makes sense to me if we need to have a default.  I was making an assumption based on the existing description of those config properties in the existing code, all of which say something like `Defaults to CIEL value set: XYZ`.  If they weren't actually defaulting to CIEL value sets, and this fixes that, that makes sense.

mseaton:
All this said, it would be nice if we could eliminate use of UUIDs like this entirely, and switch to mappings.

mogoodrich (approved):
Yeah, the intent here (as noted in the property descriptions) is for these default values to point to the default values provided by CIEL.  This seems correct to me.  Likely a bug/oversight on my part that these were set to PIH-specific uuids.

@denniskigen  the reason (I believe) that "close" and "pause" point to the same concept is because while there was a request in our use case to have different reason sets (which we do in the PIH dictionary) the CIEL dictionary only had a single set.

So this LGTM!

denniskigen:
Merging this in shortly. Thanks all.

PR #154, opened 2025-03-19
(feat):O3-4547 Enhance prescriptions printout to include prescriber

Omoshlawi:
## Requirements

## Screenshots

Screenshot 2025-03-19 at 08 40 07

## Related Issue

https://openmrs.atlassian.net/browse/O3-4547

### DIFF
--- a/src/print-prescription/prescription-printout.component.tsx
+++ b/src/print-prescription/prescription-printout.component.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
 import { Layer, StructuredListBody, StructuredListCell, StructuredListRow, StructuredListWrapper } from '@carbon/react';
@@ -24,6 +24,7 @@ const PrescriptionsPrintout: React.FC<PrescriptionsPrintoutProps> = ({ medicatio
     sessionLocation: { display: facilityName },
   } = useSession();
   const patient = medicationrequests[0]?.request?.subject;
+  const reqester = useRef<string>(null);
 
   const extractPatientName = (display: string) => (display.includes('(') ? display.split('(')[0] : display);
   return (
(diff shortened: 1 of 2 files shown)


src/print-prescription/prescription-printout.component.tsx
```diff
@@ -118,6 +120,11 @@ const PrescriptionsPrintout: React.FC<PrescriptionsPrintoutProps> = ({ medicatio
                 </div>
               );
             })}
+          {reqester.current && (
+            <p className={styles.prescriber}>
+              {t('prescribedBy', 'Prescribed By')}: {reqester.current?.split('(')?.at(0)}
```
mogoodrich, line 125: Just to clarify, what is the "split" doing here?

src/print-prescription/prescription-printout.component.tsx
```diff
@@ -24,6 +24,7 @@ const PrescriptionsPrintout: React.FC<PrescriptionsPrintoutProps> = ({ medicatio
     sessionLocation: { display: facilityName },
   } = useSession();
   const patient = medicationrequests[0]?.request?.subject;
+  const reqester = useRef<string>(null);
```
mogoodrich, line 27: Typo, should be "requester".

Also, what is the reasoning for using a "useRef" here, is it needed?

mogoodrich (commented):
Generally looks good @Omoshlawi , just a few questions, thanks!

src/print-prescription/prescription-printout.component.tsx
Omoshlawi: the requester name has username, so i was stripping off the username e.g. `Sarah Linda Foster (Identifier: admin) `

mogoodrich: Makes sense, thanks @Omoshlawi !

src/print-prescription/prescription-printout.component.tsx
mogoodrich: Ah, I think I understand now... looking at the code, however, it looks like if, for some reason, the prescriptions have *different* prescriber, it won't show all of them?

Also in most cases there is only one prescriber, I would still prefer we display a comma-separated list of all the providers in the case that there are multiple.  Similar to what we did here (assigning them to a Set makes sure we don't display duplicates):

https://github.com/openmrs/openmrs-esm-dispensing-app/blob/main/src/medication-request/medication-request.resource.tsx#L120

Does this make sense? Can you look into making this change?  Thanks!

Omoshlawi: Yeah, makes alot of sense, and I have fixed it.Thanks

mogoodrich (approved):
LGTM, thanks @Omoshlawi !

jwnasambu:
@Omoshlawi Thanks for working on this issue.

PR #155, opened 2025-03-27
(feat) O3-4571 : Make batch validity check configurable

jabahum:
## Requirements

## Summary

This PR disables batch number validity check.
Before
image

After
image

## Related Issue

https://openmrs.atlassian.net/browse/O3-4571

### DIFF
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -111,7 +111,7 @@ export const configSchema = {
     _type: Type.Boolean,
     _description:
       'Enable or disable stock deduction during the dispensing process. Requires the stock management module to be installed and configured.',
-    _default: true,
+    _default: false,
   },
 
   validateBatch: {
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -113,12 +113,11 @@ export const configSchema = {
       'Enable or disable stock deduction during the dispensing process. Requires the stock management module to be installed and configured.',
     _default: false,
   },
-
   validateBatch: {
     _type: Type.Boolean,
     _description:
-      'Enable or disable stock item batch number validation.Requires the stock management module to be installed and configured.',
-    _default: false,
+      'Enable or disable stock item batch number validation. Requires the stock management module to be installed and configured.',
+    _default: true,
   },
 };
 
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -111,6 +111,13 @@ export const configSchema = {
     _type: Type.Boolean,
     _description:
       'Enable or disable stock deduction during the dispensing process. Requires the stock management module to be installed and configured.',
+    _default: true,
+  },
+
+  validateBatch: {
+    _type: Type.Boolean,
+    _description:
+      'Enable or disable stock item batch number validation.Requires the stock management module to be installed and configured.',
     _default: false,
   },
 };
@@ -156,4 +163,6 @@ export interface PharmacyConfig {
     };
   };
   enableStockDispense: boolean;
+
+  validateBatch: boolean;
 }
--- a/src/forms/stock-dispense/stock-dispense.component.tsx
+++ b/src/forms/stock-dispense/stock-dispense.component.tsx
@@ -2,8 +2,9 @@ import React from 'react';
 import { ComboBox, InlineLoading, InlineNotification, Layer } from '@carbon/react';
 import { type MedicationDispense, type InventoryItem } from '../../types';
 import { useDispenseStock } from './stock.resource';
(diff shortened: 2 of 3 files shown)


makombe (approved):
Thanks @jabahum  LGTM!

src/config-schema.ts
```diff
@@ -111,6 +111,13 @@ export const configSchema = {
     _type: Type.Boolean,
     _description:
       'Enable or disable stock deduction during the dispensing process. Requires the stock management module to be installed and configured.',
+    _default: true,
```
makombe, line 114: Should this be false by default?

jabahum: This has been resolved

ibacher:
I marked this as `(feat)` because this is actually a user-visible change. I don't see how the changes here relate to the _batch number_, though. It looks like the validity check is entirely to do with expiration dates?

Also, in the ticket, can you please describe the actual issue you're encountering. Slack is not a long-term reference for what the issue is.

ibacher (changes requested):
See my previous questions. This mostly goes to how we're naming and describing things.

jabahum:
> I marked this as `(feat)` because this is actually a user-visible change. I don't see how the changes here relate to the _batch number_, though. It looks like the validity check is entirely to do with expiration dates?
> 
> Also, in the ticket, can you please describe the actual issue you're encountering. Slack is not a long-term reference for what the issue is.

The changes here are actually just disabling the use of the validity check which check for the format of the batchnumuber in the items returned in the hook

jabahum:
> I marked this as `(feat)` because this is actually a user-visible change. I don't see how the changes here relate to the _batch number_, though. It looks like the validity check is entirely to do with expiration dates?
> 
> Also, in the ticket, can you please describe the actual issue you're encountering. Slack is not a long-term reference for what the issue is.

i have also provided a bit of description on the issue to complement  on the ticket  on jira

jabahum:
@ibacher any thoughts on how this could be done better.

ibacher:
@jabahum The main point is I just don't understand the relationship between "batch number" and [the function you're disabling](https://github.com/openmrs/openmrs-esm-dispensing-app/blob/ba73800993ad8e55efb0d1414368b66a58deeb33/src/forms/stock-dispense/stock-dispense.component.tsx#L36). The reason this is a concern is because you keep referencing batch number, e.g., in the description of the configuration option (configurations are a user interface and this is why I described this as a user-facing change), but the check doesn't seem to have anything to do with batch numbers.

But that said, I would move the configuration check into `isValidBatch()` and just have it return `true` if it's disabled.

src/config-schema.ts
```diff
@@ -113,6 +113,13 @@ export const configSchema = {
       'Enable or disable stock deduction during the dispensing process. Requires the stock management module to be installed and configured.',
     _default: false,
   },
+
+  validateBatch: {
+    _type: Type.Boolean,
+    _description:
+      'Enable or disable stock item batch number validation.Requires the stock management module to be installed and configured.',
```
ibacher, line 120:
```suggestion
      'Enable or disable stock item batch number validation. Requires the stock management module to be installed and configured.',
```

src/config-schema.ts
```diff
@@ -156,4 +163,6 @@ export interface PharmacyConfig {
     };
   };
   enableStockDispense: boolean;
+
```
ibacher, line 166: Why the empty line here?

src/config-schema.ts
```diff
@@ -113,6 +113,13 @@ export const configSchema = {
       'Enable or disable stock deduction during the dispensing process. Requires the stock management module to be installed and configured.',
     _default: false,
   },
+
+  validateBatch: {
+    _type: Type.Boolean,
+    _description:
+      'Enable or disable stock item batch number validation.Requires the stock management module to be installed and configured.',
+    _default: false,
```
ibacher, line 121: This seems like `_default: true` is the right way of keeping the current behaviour by default?

src/forms/stock-dispense/stock-dispense.component.tsx
```diff
@@ -31,6 +34,7 @@ const StockDispense: React.FC<StockDispenseProps> = ({ medicationDispense, updat
 
   //check whether the drug will expire before the medication period ends
   function isValidBatch(medicationToDispense, inventoryItem) {
+    if (!config?.validateBatch) return true;
```
ibacher, line 37:
```suggestion
    if (typeof config !== 'undefined' && !config.validateBatch) {
       return true;
    }
```

ibacher (approved):

PR #418, opened 2025-08-12
(fix) O3-4956: Completed medication order should update the fulfiller status

NethmiRodrigo:
## Requirements

## Summary

This PR updates the computeFulfillerStatus function to return `Completed` if the medication dispense has the status completed. This makes it so that the status column on the patient chart's order column gets updated with the status when a medication order is dispensed.

## Other

The reason this is a draft is because I'm a bit confused. Right now, the `fulfillerStatus` is set to `Completed` only if the config `restrictTotalQuantityDispensed` is set to true and the total dispensing quantity amount has been met.  Right now, this means that looking at the order's table on the patient chart, you wouldn't be able to tell if an order has been dispensed (even if it has been dispensed, multiple times). 
I've made a change so that if the most recent medication dispense's status is completed, then we set the `fulfillerStatus` to completed too, but I suspect that might not be the right thing to do here.

### DIFF
--- a/src/utils.test.ts
+++ b/src/utils.test.ts
@@ -850,7 +850,6 @@ describe('Util Tests', () => {
       newMedicationDispense.extension[0].valueDateTime = '2023-01-03T14:00:00-05:00';
       newMedicationDispense.whenHandedOver = '2023-01-03T14:00:00-05:00';
       newMedicationDispense.whenPrepared = '2023-01-03T14:00:00-05:00';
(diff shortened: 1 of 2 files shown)


denniskigen:
@NethmiRodrigo Remember to close this out (and provide an explanation why).

NethmiRodrigo:
Closing this PR since the underlying purpose of this PR was to have some sort of way to complete an order which was achieved by #420

PR #428, opened 2025-08-17
(fix) O3-4692: Display "Dispensed" status in Prescription Details section

dilankavishka:
## Requirements

## Summary

This PR implements the display of "Dispensed" status in the Prescription Details section to ensure consistency with other statuses like "Cancelled" and "Paused". Previously, the "Dispensed" status only appeared in the History and Comments section, which created an inconsistent user experience.

## Screenshots
https://github.com/user-attachments/assets/8f5bd5f3-a018-42b9-8272-1ad43f21fadf

## Related Issue
https://openmrs.atlassian.net/browse/O3-4692

## Other
N/A


dilankavishka:
Hi @VeronicaMuthee , @denniskigen  Could you review this PR please?

dilankavishka:
@mogoodrich @NethmiRodrigo @VeronicaMuthee Could you review this PR please?

src/types.ts
```diff
@@ -376,6 +376,7 @@ export enum MedicationRequestCombinedStatus {
   on_hold = 'on-hold',
   declined = 'declined',
   unknown = 'unknown',
+  dispensed = 'dispensed',
```
mogoodrich: The problem here is that "dispensed" and "active" are overlapping states--ie, any order that is "dispensed" is also "active"

src/utils.ts
```diff
@@ -80,14 +80,16 @@ export function computeFulfillerStatus(
  * @param medicationRequestExpirationPeriodInDays
  */
 export function computeMedicationRequestCombinedStatus(
```
mogoodrich, line 82: I'm not sure we want to change this method, for the reasons I mentioned above: "active" and "dispensed" are kind of overlapping statuses so this is a little confusing.

mogoodrich (changes requested):
Hmmm.... this becomes a little trickier than I thought.  I don't think we want to add another state, I'm wondering if we can restrict this change to just the "generateStatusTag"... that is, have the generateStatusTag call computeMedicationRequestCombinedStatus, as it does, but if that method returns null, check if there are any dispenses (within the generateStatusTag method) and turn the new "Dispensed" tag if there are.  Thoughts?

Let me ask one question on the thread as well.

src/types.ts
dilankavishka: fixed by handle the "Dispensed" tag solely in the UI presentation layer

src/utils.ts
dilankavishka: The overall prescription will correctly remain as 'Active', and the logic for displaying the 'Dispensed' tag next to the drug has been moved entirely into the `generateStatusTag` method. I've Updated the PR description with the screen record

src/prescriptions/prescription-details.component.tsx
```diff
@@ -28,11 +38,16 @@ const PrescriptionDetails: React.FC<{
...
     const combinedStatus: MedicationRequestCombinedStatus = computeMedicationRequestCombinedStatus(
-      medicationRequest,
+      medicationRequestBundle.request,
       config.medicationRequestExpirationPeriodInDays,
     );
+
+    const mostRecentDispenseStatus = getMostRecentMedicationDispenseStatus(medicationRequestBundle.dispenses);
+    if (mostRecentDispenseStatus === MedicationDispenseStatus.completed) {
+      return <Tag type="gray">{t('dispensed', 'Dispensed')}</Tag>;
+    }
```
mogoodrich: Let's move this to be the last check that happens, that is that this check should be after the "on-hold" check--only if there is no "combined status" should we consider displaying "dispensed".   And let's add a comment "// if there is not combined status, but the last event was a dispense, display "dispensed"

mogoodrich (changes requested):
Generally looks good @dilankavishka , just one change.

src/prescriptions/prescription-details.component.tsx
dilankavishka: Thank you so much @mogoodrich ! and I fixed this

mogoodrich (approved):
Thanks @dilankavishka ! Looks good to me!  I will make sure the test pass and then merge.

mogoodrich (changes requested):
Oh @dilankavishka  it looks like there is an E2E failure... it looks like the test just needs to be updated for the change in functionality, can you take a look?

dilankavishka:
Yes! @mogoodrich. I think the main reason is now we have two "Dispensed" tags and it threw the strict mode violation. 

Updated the test

mogoodrich:
This looks right @dilankavishka , but if it is easy enough do we want to test that the new "Dispensed" you added is also present as well?

dilankavishka:
Definitely @mogoodrich. I updated the e2e test to see 'Dispensed' tag is present within the Prescription details section as well.

dilankavishka:
Tests are passing. :)

Thank you so much for the support @mogoodrich !!!

mogoodrich:
Merging, thanks @dilankavishka !

mogoodrich (approved):

PR #435, opened 2025-08-19
(fix) O3-4982:  Cancel Disabled while submitting a form

jwnasambu:
## Requirements

## Summary

In the Close, Pause, Dispense components, clicking the Save button disables the Cancel button because both buttons relied on the isSubmitting state. I therefore removed the disabled={isSubmitting} prop from the Cancel button, so it remains clickable during submission.

## Screenshots

**Before making changes:**

https://github.com/user-attachments/assets/627853ba-95f3-4aa4-9eeb-88d958d45486

**After fixing the issue:** 

https://github.com/user-attachments/assets/671b22ff-2e7c-4f13-8fa3-66b7ee9eddd0

## Related Issue

https://openmrs.atlassian.net/browse/O3-4982

### DIFF
--- a/src/forms/close-dispense-form.workspace.tsx
+++ b/src/forms/close-dispense-form.workspace.tsx
@@ -176,7 +176,7 @@ const CloseDispenseForm: React.FC<CloseDispenseFormProps> = ({
         </section>
       </div>
       <section className={styles.buttonGroup}>
-        <Button disabled={isSubmitting} onClick={() => closeWorkspace()} kind="secondary">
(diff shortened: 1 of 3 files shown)


mogoodrich (commented):
Thanks @jwnasambu !

But we don't really *want* the cancel button to be clickable while we are submitting the form, do we?  If the user does hit the cancel button in the midst of submission is the behavior consistent and what we want?  fyi @denniskigen

denniskigen:
Inclined to agree with @mogoodrich here.

PR #444, opened 2025-08-21
(fix) O3-4984 Display Today's Date instead of "Today" in dispensing app

ELVIS-KATO:
## Requirements

## Summary

Changed the option: From { mode: 'standard' } to { noToday: true }
The noToday: true option tells the formatDate function to always display the actual date instead of converting it to relative text like "today". Just like it has always been
## BEFORE:
Todays_Date Before
 
## THEN AFTER:
Todays_Date After

## Related Issue

[https://openmrs.atlassian.net/issues/?filter=-4&selectedIssue=O3-4984](https://openmrs.atlassian.net/issues/?filter=-4&selectedIssue=O3-4984)

## Other

@fiona855 ,

### DIFF
--- a/src/pharmacy-header/pharmacy-header.component.tsx
+++ b/src/pharmacy-header/pharmacy-header.component.tsx
@@ -27,7 +27,7 @@ export const PharmacyHeader: React.FC = () => {
           <span className={styles.value}>{userLocation}</span>
           <span className={styles.middot}>&middot;</span>
           <Calendar size={16} />
-          <span className={styles.value}>{formatDate(new Date(), { mode: 'standard' })}</span>
+          <span className={styles.value}>{formatDate(new Date(), { noToday: true })}</span>
         </div>
       </div>
     </div>


src/pharmacy-header/pharmacy-header.component.tsx
```diff
@@ -27,7 +27,7 @@ export const PharmacyHeader: React.FC = () => {
           <span className={styles.value}>{userLocation}</span>
           <span className={styles.middot}>&middot;</span>
           <Calendar size={16} />
-          <span className={styles.value}>{formatDate(new Date(), { mode: 'standard' })}</span>
+          <span className={styles.value}>{formatDate(new Date(), { noToday: true })}</span>
```
mogoodrich, line 30: Not sure what mode "standard" is, but do we still want it? @denniskigen 

Otherwise, looks good, thanks @ELVISKATS

mogoodrich (approved):

src/pharmacy-header/pharmacy-header.component.tsx
denniskigen: `standard` mode is the default, so it's implicitly applied:

```
'standard': "22-Aug-2025" (compact)
'wide': "22 — Aug — 2025" (with separators)
```

denniskigen (approved):

PR #482, opened 2025-09-10
(fix) O3-5018 prevent user from dispensing an already dispensed presc…

chibongho:
…ription

## Requirements

## Summary

In this PR, we disable the Cancel / Pause / Dispense buttons on a prescription when the user has already initiated a request to do one of those actions, and we only re-enable the buttons after the request has completed AND the encounter data has revalidated. This prevents users with slower connections to be able to click on the Dispense button while the revalidatation has not completed.

## Screenshots

See video. Notice that the Cancel / Pause / Dispense buttons (partially covered by the workspace) are disabled as soon as we click the "Dispense" button. 
https://github.com/user-attachments/assets/72b8b080-17dc-452e-a537-d3f0d726c99e

## Related Issue

https://openmrs.atlassian.net/browse/O3-5018

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3071,9 +3071,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@openmrs/esm-api@npm:8.0.1-pre.3305":
-  version: 8.0.1-pre.3305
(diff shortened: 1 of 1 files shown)


mogoodrich (approved):
Awesome, looks great, thanks @chibongho !

PR #489, opened 2025-09-15
(feat) O3-5005 add search tab to dispensing home page

chibongho:
## Requirements

## Summary

This PR adds a search tab in the dispensing home page. Functionally, this tab is really similar to the "Active prescription" tab. The only difference is that it does not initiate request for data until the user searches for user.

Other changes:
- Refactored out the prescriptions table to it's re-usable across all the tabs
- The "Active prescription" and "All prescription" tabs still have a search bar. However, previously the search bar is shared across those 2 tabs. With this change, each tab has its own search bar.
- The wording in the new search tab deviates slightly from this (old?) [mockup](https://app.zeplin.io/project/6650717a9275db731847fb32/screen/66585cb0ed5b6e2b45b330eb) for better consistency with the other apps.

Marking this as draft as I'm using a placeholder icon in lieu of the real search icon in pending PR https://github.com/openmrs/openmrs-esm-core/pull/1454. Code-wise, this is ready for review.

## Screenshots
See video. Note that patient "Manu" is searched twice, once via pressing the "Enter" key and once by clicking on the "Search" button.

https://github.com/user-attachments/assets/cc8d3536-a65c-4880-a787-a95264b9e9a0

## Related Issue

https://openmrs.atlassian.net/browse/O3-5005


src/prescriptions/patient-search-tab-panel.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+            status={'ACTIVE'}
+            debouncedSearchTerm={submittedSearchTerm}
+            location={''}
+          />
+        ) : (
+          <div className={styles.searchForPatientPlaceholder}>
+            <div>
+              <AppointmentsPictogram />
+              <h5>Search for a patient</h5>
+              <div>Search for a patient by name or identifier number</div>
```
pirupius, line 49: I think you forgot to include the translation strings for these

mogoodrich (approved):
Generally LGTM, though we do need to add the translations @pirupius  pointed out.

I didn't see the location filter in the screenshot... is that going away?

donaldkibet:
Thanks @chibongho, LGTM

This might be slightly beyond the scope of this PR, but could we consider using a stateful URL for filtering **Search**, **Active Prescription**, and **All Prescription**?

For example, if we had a URL like:
```
dispensing?patientUuid=some-patient-uuid
```
it would filter prescriptions for that patient across all tabs.

The main purpose is to integrate better with queues — when a patient is queued for pharmacy, the pharmacist shouldn’t need to search again to view their active prescriptions.

PR #512, opened 2025-09-29
(fix) O3-5019: Fix infinite re-renders in dispensing workspace

chibongho:
## Requirements

## Summary

When opening the dispensing form, the form enters an infinite re-rendering loop. See Screenshot below. This PR addresses the this bug and other minor performance issues (like avoid using React states for derived values)

## Screenshots

Before:
https://github.com/user-attachments/assets/e7122850-846b-4638-b338-9c98300e62d3

After, (ignore the crash at the end from [mismatched dispensing units]( https://openmrs.atlassian.net/browse/O3-5019)):
https://github.com/user-attachments/assets/392068bd-9e5e-4279-892a-c6f9f133cccb

## Related Issue

https://openmrs.atlassian.net/browse/O3-5019

### DIFF
--- a/src/forms/medication-dispense-review.component.tsx
+++ b/src/forms/medication-dispense-review.component.tsx
@@ -7,6 +7,7 @@ import { getConceptCodingUuid, getMedicationReferenceOrCodeableConcept, getOpenM
 import { useMedicationCodeableConcept, useMedicationFormulations } from '../medication/medication.resource';
 import { useMedicationRequest, usePrescriptionDetails } from '../medication-request/medication-request.resource';
 import {
+  blankSubstitution,
   useOrderConfig,
   useProviders,
   useSubstitutionReasonValueSet,
@@ -139,23 +140,21 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
     medicationRequest?.medicationReference?.reference &&
     medicationDispense?.medicationReference?.reference &&
     medicationRequest.medicationReference.reference != medicationDispense.medicationReference.reference;
-  const substitution = useMemo(() => {
+
+  const { substitution } = medicationDispense;
+  useEffect(() => {
     if (isSubstitution) {
-      // make sure that the value medicationDispense.substitution.wasSubstituted exists and is truthy
-      if (medicationDispense.substitution.wasSubstituted) {
-        return medicationDispense.substitution;
-      } else {
-        return { ...medicationDispense.substitution, wasSubstituted: true };
+      // make sure that the value substitution.wasSubstituted exists and is truthy
+      if (!substitution.wasSubstituted) {
+        const newSubstitution = { ...substitution, wasSubstituted: true };
+        updateMedicationDispense({ substitution: newSubstitution });
       }
     } else {
-      return blankSubstitution;
+      if (substitution != blankSubstitution) {
+        updateMedicationDispense({ substitution: blankSubstitution });
+      }
     }
-  }, [isSubstitution, medicationDispense]);
-  useEffect(() => {
-    updateMedicationDispense({
-      substitution,
-    });
-  }, [substitution, updateMedicationDispense]);
+  }, [isSubstitution, substitution, updateMedicationDispense]);
 
   useEffect(() => {
     setUserCanModify(session?.user && userHasAccess(PRIVILEGE_CREATE_DISPENSE_MODIFY_DETAILS, session.user));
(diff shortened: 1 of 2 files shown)


src/forms/medication-dispense-review.component.tsx
```diff
@@ -20,7 +20,7 @@ import styles from '../components/medication-dispense-review.scss';
 
 interface MedicationDispenseReviewProps {
   medicationDispense: MedicationDispense;
-  updateMedicationDispense: Function;
+  updateMedicationDispense(updates: Partial<MedicationDispense>);
```
chibongho, line 23: this function was made to accept a `Partial<MedicationDispense>` instead of the full `MedicationDispense` object. This removes the need to put the `medicationDispense` object in the dependency arrays of a few `useEffect` calls, which were the main source of the infinite re-render.

src/forms/medication-dispense-review.component.tsx
```diff
@@ -164,36 +135,18 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
   );
 
   // check to see if the current dispense would be a substitution, and update accordingly
+  const isSubstitution =
+    medicationRequest?.medicationReference?.reference &&
+    medicationDispense?.medicationReference?.reference &&
+    medicationRequest.medicationReference.reference != medicationDispense.medicationReference.reference;
+  const substitution = useMemo(() => {
+    return isSubstitution ? medicationDispense.substitution : blankSubstitution;
+  }, [isSubstitution, medicationDispense]);
```
denniskigen, line 144: When a user selects a different medication (creating a substitution), the new code doesn't set `wasSubstituted: true`:

When `isSubstitution` is true, it returns `medicationDispense.substitution` as-is, without setting `wasSubstituted: true`. This means:
  - User changes medication A → B (creating substitution)
  - `isSubstitution` becomes true
  - But `medicationDispense.substitution.wasSubstituted` remains false
  - Validation passes WITHOUT requiring substitution reason/type (because !wasSubstituted is truthy)
  - This bypasses required substitution documentation

The old code explicitly set this flag:

```tsx
  substitution: {
    ...medicationDispense.substitution,
    wasSubstituted: true,  // explicitly set
  },
```
 
The change below ensures `wasSubstituted` accurately reflects whether medications differ, which then triggers the [validation](https://github.com/openmrs/openmrs-esm-dispensing-app/blob/f4642d5cd29afa14ba204975379a2ffdbfaa6d8b/src/forms/dispense-form.workspace.tsx#L195) requiring reason/type fields.

```suggestion
  const substitution = useMemo(() => {
    return isSubstitution
      ? { ...medicationDispense.substitution, wasSubstituted: true }
      : blankSubstitution;
  }, [isSubstitution, medicationDispense]);
```

src/forms/medication-dispense-review.component.tsx
```diff
@@ -164,36 +135,27 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
...
-        ...medicationDispense,
-        substitution: {
-          ...medicationDispense.substitution,
-          wasSubstituted: true,
-        },
-      });
+  const isSubstitution =
+    medicationRequest?.medicationReference?.reference &&
+    medicationDispense?.medicationReference?.reference &&
+    medicationRequest.medicationReference.reference != medicationDispense.medicationReference.reference;
```
denniskigen, line 141:
```suggestion
  const isSubstitution = useMemo(
    () =>
      medicationRequest?.medicationReference?.reference &&
      medicationDispense?.medicationReference?.reference &&
      medicationRequest.medicationReference.reference != medicationDispense.medicationReference.reference,
    [
      medicationRequest?.medicationReference?.reference,
      medicationDispense?.medicationReference?.reference,
    ],
  );
  
```

nit: Minor thing, but we could consider memoizing `isSubstitution` so it's not recalculated on every render. That said, it's not the most complex computation, hence the nit prefix.

src/forms/medication-dispense-review.component.tsx
```diff
@@ -164,36 +135,27 @@ const MedicationDispenseReview: React.FC<MedicationDispenseReviewProps> = ({
...
-              coding: [{ code: null }],
-            },
-          ],
-          type: { coding: [{ code: null }] },
-        },
-      });
+      return blankSubstitution;
     }
-  }, [medicationDispense, medicationRequest, updateMedicationDispense]);
+  }, [isSubstitution, medicationDispense]);
```
denniskigen, line 153:
```suggestion
  }, [isSubstitution, medicationDispense.substitution]);
  
```

Can we change the substitution deps  in the deps array from `medicationDispense` to `medicationDispense.substitution`? This should prevent recalculation when quantity/dosage changes.

denniskigen (commented):
Two more things I didn't catch in the original review, @chibongho.

PR #743, opened 2026-02-04
(feat) O3-5401: Show only drug order panels in the order basket

Muppasanipraneeth:
## Requirements

- [x] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide) and [**design documentation**](https://zeroheight.com/23a080e38/p/880723-introduction).

## Summary

When the dispensing app launches the order basket to fill a prescription, all order panels (medications, labs, etc.) are shown. This is confusing — a pharmacist filling a prescription shouldn't see unrelated order types.

This PR passes `visibleOrderPanels: [drugOrderTypeUUID]` when launching the order basket, so only the drug order panel is displayed. This uses the filtering support added in openmrs/openmrs-esm-patient-chart#2990.

A new `drugOrderTypeUUID` config property is added to the dispensing app's config schema. Its default matches the medications app's `orderTypeUuid` config. Deployments that override the drug order type UUID in the medications app must also override it here.

### Before

https://github.com/user-attachments/assets/3e94dd0b-2945-429c-9581-f6e95526839f

### After

https://github.com/user-attachments/assets/72853ebe-0801-4805-ac19-b6c337f53c18

## Related Issue

https://openmrs.atlassian.net/browse/O3-5401

## Other

Related PRs:
- openmrs/openmrs-esm-patient-chart#2990 (adds `visibleOrderPanels` support)
- openmrs/openmrs-esm-laboratory-app#661 (lab app consumer)


VeronicaMuthee:
@Muppasanipraneeth this looks good to me, thank you.

RajPrakash681:
@VeronicaMuthee The changes look clean . I tested this locally and everything works as expected. @Muppasanipraneeth

src/config-schema.ts
```diff
@@ -2,6 +2,11 @@ import { Type, validators } from '@openmrs/esm-framework';
 import { type CustomTab } from './types';
 
 export const configSchema = {
+  drugOrderTypeUUID: {
+    _type: Type.UUID,
+    _description: "UUID for the 'Drug' order type to fetch medications",
+    _default: '131168f4-15f5-102d-96e4-000c29c2a5d7',
```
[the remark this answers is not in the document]
chibongho: Its also a configuration in the esm-patient-chart app. We can theoretically pull it from there, assuming that the patient-chart app is installed, with: `useConfig({ externalModuleName: '@openmrs/esm-patient-chart-app' })`

denniskigen:
I agree that keeping a local config is the safer option here. Using `useConfig({ externalModuleName: '@openmrs/esm-patient-medications-app' })` in this component isn't a clean fallback, because `useConfig` suspends until the external module's config store is loaded rather than returning `undefined`/`{}` immediately.

The one thing I'd still tighten is the config description. This UUID is not being used to "fetch medications"; it's being passed to `visibleOrderPanels`, which filters the order basket against each extension's `config.orderTypeUuid`. So the description should make that coupling explicit, e.g.:

```ts
_description: "UUID for the 'Drug' order type, used to filter the order basket to show only drug order panels. Must match the orderTypeUuid config in @openmrs/esm-patient-medications-app.",
```

With that clarification, this looks like a clean consumer of the `visibleOrderPanels` support added in [openmrs-esm-patient-chart#2990.](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2990).

e2e/commands/drug-order-operations.ts
```diff
@@ -1,16 +1,18 @@
...
 import { type Encounter } from './types';
+import type { PharmacyConfig } from '../../src/config-schema';
 
 export const generateRandomDrugOrder = async (
   api: APIRequestContext,
   patientId: string,
   encounter: Encounter,
   providerUuid: string,
 ): Promise<Order> => {
+  const { drugOrderTypeUUID } = useConfig<PharmacyConfig>();
```
denniskigen, line 12: `useConfig` is a React hook, but this helper runs in the Playwright/Node test process, outside any React render or OpenMRS `ComponentContext`. It will throw "Invalid hook call" before `api.post` runs, and every spec that calls `generateRandomDrugOrder` in `beforeEach` (`active-prescriptions`, `close-prescription`, `dispense-medication`, `pause-prescription`) will fail before creating its order. Keep this as a test constant / env value, or pass the UUID in from the test setup.

src/config-schema.ts
```diff
@@ -2,6 +2,11 @@ import { Type, validators } from '@openmrs/esm-framework';
 import { type CustomTab } from './types';
 
 export const configSchema = {
+  drugOrderTypeUUID: {
+    _type: Type.UUID,
+    _description: "UUID for the 'Drug' order type, used to filter the order basket to show only drug order panels.Must match the orderTypeUuid config in @openmrs/esm-patient-medications-app.",
```
denniskigen, line 7: Missing space: `panels.Must` → `panels. Must`.

denniskigen (dismissed):
Thanks for iterating on this, @Muppasanipraneeth! The runtime change is the right shape, but the e2e update won't run. Details inline.

mogoodrich:
> The one thing I'd still tighten is the config description. This UUID is not being used to "fetch medications"; it's being passed to `visibleOrderPanels`, which filters the order basket against each extension's `config.orderTypeUuid`. So the description should make that coupling explicit, e.g.:

Is the issue here is that we don't have a shared config somewhere where this value can be pulled from, and that's why we need this new config property?

chibongho:
> > The one thing I'd still tighten is the config description. This UUID is not being used to "fetch medications"; it's being passed to `visibleOrderPanels`, which filters the order basket against each extension's `config.orderTypeUuid`. So the description should make that coupling explicit, e.g.:
> 
> Is the issue here is that we don't have a shared config somewhere where this value can be pulled from, and that's why we need this new config property?

Yeah, I think so. Based on what @denniskigen said, the config value from a different app might not have been loaded before being used here. Maybe it's worth thinking about having a more global config namespace that is guaranteed to load?

mogoodrich:
> Yeah, I think so. Based on what @denniskigen said, the config value from a different app might not have been loaded before being used here. Maybe it's worth thinking about having a more global config namespace that is guaranteed to load?

Definitely worth talking about. I know Mike has brought it up.  (But no need to block this ticket on it)

mogoodrich (approved):

denniskigen:
Mind weighing in @ibacher?

ibacher:
> Based on what @denniskigen said, the config value from a different app might not have been loaded before being used here.

Technically, any configuration provided in a config.json file to the framework is available in the config store prior to pretty much anything loading. The part that depends on app loading (and hence suspense) are the defaults, since those are configured in the configuration schema and the configuration schema is created by executing Javascript code.

> Maybe it's worth thinking about having a more global config namespace that is guaranteed to load?

We sort of have one of these, `@openmrs/esm-app-shell` (and `@openmrs/esm-styleguide`) are both loaded before any apps.

Aside: long-term, I want to out-source config schemas to JSON files we can package up like the routes.json files.

denniskigen (approved):
Thanks for iterating on this, @Muppasanipraneeth.

PR #751, opened 2026-02-12
(feat) O3-5497: Enable a facility location to accept multiple Associated Pharmacy Locations.

walteronoh:
## Requirements

- [ ] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide) and [**design documentation**](https://zeroheight.com/23a080e38/p/880723-introduction).

## Summary

Currently, the Associated Pharmacy Location can only accept one location. This PR enables application of several associated pharmacy locations. Given a scenario where we have a facility named KGH, it can have two or more associated pharmacies i.e, 1. KGH Revolving Pharmacy and 2. KGH Main Pharmacy. Now from KGH location, we can filter either by the current location or by the associated pharmacy locations. 
We are also introducing two config properties namely: 

1. useAssociatedPharmacyLocations - For showing medication requests from all associated pharmacy locations of the current location.
2. useCurrentLocation - For showing medication requests from the current location.

## Screenshots

a. Without current location filter:
image
b. With current location filter:
image

### DIFF
--- a/src/prescriptions/prescription-tab-panel.component.tsx
+++ b/src/prescriptions/prescription-tab-panel.component.tsx
@@ -31,10 +31,31 @@ const PrescriptionTabPanel: React.FC<PrescriptionTabPanelProps> = ({
   // set any initially selected locations
   useEffect(() => {
     if (!isInitialized.current && !isFilterLocationsLoading && sessionLocation?.uuid) {
-      setFilterLocations(locations?.filter((l) => sessionLocation?.uuid === l.associatedPharmacyLocation) || []);
+      const initialLocations: SimpleLocation[] = [];
+      // Should display from the location associated with the pharmacy location
+      initialLocations.push(
+        ...(locations?.filter((l) => l.associatedPharmacyLocations?.includes(sessionLocation?.uuid)) || []),
+      );
+
+      // Should display from all the associated pharmacy locations of the current location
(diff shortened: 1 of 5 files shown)


src/prescriptions/prescription-tab-panel.component.tsx
```diff
@@ -31,10 +31,31 @@ const PrescriptionTabPanel: React.FC<PrescriptionTabPanelProps> = ({
...
+      );
+
+      // Should display from all the associated pharmacy locations of the current location
+      if (config.locationBehavior.locationFilter.useAssociatedPharmacyLocations) {
+        initialLocations.push(
+          ...(locations?.filter((v) =>
+            locations?.find((l) => l.id === sessionLocation?.uuid).associatedPharmacyLocations?.includes(v.id),
+          ) || []),
+        );
+      }
```
mogoodrich: This is confusing and I think unneeded, because it is setting up a mapping in both directions?  The idea is that an "associatedPharmacyLocation" is applied to a location to determine the pharmacies to associated with that location, not vice versa.

mogoodrich (changes requested):
Thanks @walteronoh !  I agree with expanding to support multiple associated pharmacy locations (with a comment) but I don't think we should add the "reverse" lookup where a pharmacy location is tagged with locations associated with it.

I'm also not crazy about the "useCurrentLocation" parameter but interested in talking more.  Can we had a ticket to discuss? Thanks!

PR #812, opened 2026-04-10
(chore) Use reusable build and release GHA workflows

NethmiRodrigo:
## Requirements

- [ ] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide) and [**design documentation**](https://zeroheight.com/23a080e38/p/880723-introduction).

## Summary

Replaces inline CI job definitions in `.github/workflows/ci.yml` with calls to reusable workflows from [openmrs/openmrs-contrib-gha-workflows](https://github.com/openmrs/openmrs-contrib-gha-workflows).

- `ci.yml` now delegates to `build-frontend-module.yml@main`

## Related Issue

N/A

## Other

Follows the pattern established in https://github.com/openmrs/openmrs-esm-billing-app/pull/703.


denniskigen (commented):
Thanks, @NethmiRodrigo! The migration to reusable workflows looks good. One thing to flag before merging.

.github/workflows/ci.yml
```diff
@@ -1,97 +1,11 @@
...
-    if: ${{ github.event_name == 'push' }}
-
-    steps:
-      - name: Trigger RefApp Build
-        uses: fjogeleit/http-request-action@master
-        with:
-          url: https://ci.openmrs.org/rest/api/latest/queue/O3-BF
-          method: "POST"
-          customHeaders: '{ "Authorization": "Bearer ${{ secrets.BAMBOO_TOKEN }}" }'
+    uses: openmrs/openmrs-contrib-gha-workflows/.github/workflows/build-frontend-module.yml@main
```
denniskigen, line 12: The old workflow ran `yarn check-translations` before `yarn verify`. This catches cases where someone adds a translation key in code but forgets to update `translations/en.json`. The reusable workflow supports this via the `pre_verify_command` input, but it is not being passed here, so that check would be silently lost.

Could you add it?

```suggestion
  build:
    uses: openmrs/openmrs-contrib-gha-workflows/.github/workflows/build-frontend-module.yml@main
    with:
      pre_verify_command: "yarn check-translations"
```

ibacher (changes requested):
See my comments on the other PR

NethmiRodrigo:
@ibacher I updated this PR based on [this](https://github.com/openmrs/openmrs-contrib-gha-workflows/pull/30) PR

.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
-name: OpenMRS CI
```
ibacher, line 1: We shouldn't change the name if possible.

.github/workflows/ci.yml
```diff
@@ -9,89 +10,27 @@ on:
...
-        uses: actions/upload-artifact@v7
-        with:
-          name: dist
-          path: dist
-          overwrite: true
+    uses: openmrs/openmrs-contrib-gha-workflows/.github/workflows/build-frontend-module.yml@main
+  with:
+      pre_verify_command: "yarn check-translations"
+  
+  owasp-dependency-check:
```
ibacher: Let's skip the dependency check. I don't think it adds a lot of value right now.

.github/workflows/ci.yml
ibacher: Better still would be to just integrate this into the upstream list of verifications that are checked for...

.github/workflows/ci.yml
NethmiRodrigo: Updated

.github/workflows/ci.yml
NethmiRodrigo: Removed

.github/workflows/ci.yml
```diff
@@ -9,89 +10,20 @@ on:
...
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
-      - name: Upload Artifacts
-        uses: actions/upload-artifact@v7
-        with:
-          name: dist
-          path: dist
-          overwrite: true
+    uses: openmrs/openmrs-contrib-gha-workflows/.github/workflows/build-frontend-module.yml@main
+  with:
+      pre_verify_command: "yarn check-translations"
```
denniskigen: The `with:` block is indented at 2 spaces, putting it at the `jobs:` level instead of under `build:`. GitHub Actions parses it as a separate job named `with` with no `runs-on` or `steps`, making the workflow invalid — which is why the latest push run completed with zero jobs. actionlint confirms the same.

```suggestion
    with:
      pre_verify_command: "yarn check-translations"
```

.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: OpenMRS CI
 
 on:
+  workflow_dispatch: # temporary, for debugging
```
denniskigen: The `# temporary, for debugging` comment should be cleaned up before merge. The trigger itself is worth keeping since `release-frontend-module.yml` supports `workflow_dispatch` in its `pre_release` condition, making it useful for manually kicking off a pre-release. Just drop the comment.

.github/workflows/ci.yml
NethmiRodrigo: Fixed

.github/workflows/ci.yml
NethmiRodrigo: Updated

.github/workflows/ci.yml
```diff
@@ -9,89 +10,20 @@ on:
...
-        uses: fjogeleit/http-request-action@master
-        with:
-          url: https://ci.openmrs.org/rest/api/latest/queue/O3-BF
-          method: "POST"
-          customHeaders: '{ "Authorization": "Bearer ${{ secrets.BAMBOO_TOKEN }}" }'
+    uses: openmrs/openmrs-contrib-gha-workflows/.github/workflows/release-frontend-module.yml@main
+    secrets:
+      NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+      BOT_PAT: ${{ secrets.OMRS_BOT_GH_TOKEN }}
+
```
denniskigen: Tiny nit before merge: the file ends with a line of trailing whitespace and no terminating newline (hence the `\ No newline at end of file` marker in the diff). Suggestion drops the stray line and adds a proper EOF newline.

```suggestion
      BOT_PAT: ${{ secrets.OMRS_BOT_GH_TOKEN }}
```

denniskigen (approved):
Thanks, @NethmiRodrigo!

PR #832, opened 2026-05-13
(feat) O3-5650: Add config to limit location filter options by visit location

chibongho:
## Requirements

## Summary

The dispensing app has a config to allow for filtering the prescriptions table by location. This PR adds a additional config, `restrictToVisitLocationDescendants`. If true, the location dropdown options are restricted to locations that are descendants of the current login location's nearest ancestor tagged as a visit location.

## Screenshots

with `restrictToVisitLocationDescendants` false:
https://github.com/user-attachments/assets/635a7743-9b85-48e2-8e82-4565e88226b3

with `restrictToVisitLocationDescendants` true:
https://github.com/user-attachments/assets/11383378-288c-4a6e-8465-aeb41861571b

### DIFF
--- a/src/location/location.resource.tsx
+++ b/src/location/location.resource.tsx
@@ -1,32 +1,93 @@
 import useSWR from 'swr';
-import { type FetchResponse, openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+import {
+  type FetchResponse,
+  type Location,
+  openmrsFetch,
+  restBaseUrl,
+  useFeatureFlag,
+  useSession,
+} from '@openmrs/esm-framework';
 import { type SimpleLocation } from '../types';
 import { type PharmacyConfig } from '../config-schema';
 import { useMemo } from 'react';
 
+export class MissingOptionalBackendDependencyError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'MissingOptionalBackendDependencyError';
+  }
+}
+
+/**
+ * Returns the list of locations to show in the location filter dropdown, based on the configuration. If the
+ * restrictToVisitLocationDescendants option is enabled, only returns locations that are descendants of the current login
+ * location's nearest ancestor tagged as a visit location.
+ * Location options are further filtered with the tag specified in the configuration.
+ */
 export function useLocations(config: PharmacyConfig) {
-  const { data, error } = useSWR<FetchResponse, Error>(
-    `${restBaseUrl}/location?tag=${encodeURIComponent(config.locationBehavior.locationFilter.tag)}&v=custom:(uuid,name,attributes:(attributeType:(name),value:(uuid))`,
-    openmrsFetch,
+  const { restrictToVisitLocationDescendants } = config.locationBehavior.locationFilter;
+  const byTag = useLocationsByTag(config, !restrictToVisitLocationDescendants);
+  const byVisit = useVisitLocationDescendants(config, restrictToVisitLocationDescendants);
+  return restrictToVisitLocationDescendants ? byVisit : byTag;
+}
+
+function toSimpleLocation(associatedPharmacyLocationAttribute: string) {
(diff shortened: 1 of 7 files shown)


src/location/location.resource.tsx
```diff
@@ -1,32 +1,93 @@
...
+function useVisitLocationDescendants(config: PharmacyConfig, enabled: boolean) {
+  const { sessionLocation } = useSession();
+  const { tag, associatedPharmacyLocationAttribute } = config.locationBehavior.locationFilter;
+  const isEMRAPIInstalled = useFeatureFlag('emrapi-module');
+
+  const configError = useMemo(
+    () =>
+      enabled && !isEMRAPIInstalled
+        ? new MissingOptionalBackendDependencyError(
+            'The restrictToVisitLocationDescendants option is enabled in the configuration, but the EMR API module is not installed.',
```
ibacher, line 70: It would be nice if we could make this message translatable, since it can be displayed to users.

ibacher (approved):

src/location/location.resource.tsx
```diff
@@ -1,32 +1,93 @@
...
+function toSimpleLocation(associatedPharmacyLocationAttribute: string) {
+  return (e: Location): SimpleLocation => ({
+    id: e.uuid,
+    name: e.name,
+    associatedPharmacyLocation:
+      e.attributes?.find((a) => a.attributeType.name === associatedPharmacyLocationAttribute)?.value?.uuid ?? null,
+  });
+}
+
+function useLocationsByTag(config: PharmacyConfig, enabled: boolean) {
```
mogoodrich, line 43: This

src/location/location.resource.tsx
```diff
@@ -1,32 +1,93 @@
...
+  const { tag, associatedPharmacyLocationAttribute } = config.locationBehavior.locationFilter;
+  const url = enabled
```
mogoodrich, line 45: This feels kind of weird to me, but maybe it's the right way to do it in react-ful hook-friendly way?  If we are enabled to restrict by visit location we call this method but then it effectively doesn't do anything because it calls the useSWR with a null value for the url?

mogoodrich (approved):
Thanks @chibongho !  One question but generally LGTM!

src/location/location.resource.tsx
chibongho: Yeah, this is awkward, but I don't think there a way around it. React hook calls cannot be conditioned (like `if(X) useYHook()` ) because on every render component must call its hooks in the exact same order. The way useSWR gets around it is by allowing for disabling the data fetching by passing in `null` as the URL.

denniskigen:
One thought on the error UX: `PrescriptionTabPanel` is rendered 3× in `prescription-tab-lists.component.tsx`, so a misconfiguration will fire three identical error notifications on first load. The description is also `locationsError.message`, which is hard-coded English and won't go through `t()`.

A few options:
- Hoist the misconfig check up to `PrescriptionTabLists` so it fires once, or surface it as an inline empty-state in the dropdown instead of a global notification.
- Use a translatable user-facing description rather than the raw `Error.message`.
- `showSnackbar` is a better fit than `showNotification` for transient feedback

chibongho:
> One thought on the error UX: `PrescriptionTabPanel` is rendered 3× in `prescription-tab-lists.component.tsx`, so a misconfiguration will fire three identical error notifications on first load. The description is also `locationsError.message`, which is hard-coded English and won't go through `t()`.
> 
> A few options:
> 
>     * Hoist the misconfig check up to `PrescriptionTabLists` so it fires once, or surface it as an inline empty-state in the dropdown instead of a global notification.
> 
>     * Use a translatable user-facing description rather than the raw `Error.message`.
> 
>     * `showSnackbar` is a better fit than `showNotification` for transient feedback

Changed it to snarkbar, Lifted it to the parent component `PrescriptionTabLists` and added translation string for it.

PR #833, opened 2026-05-14
(feat) O3-5690: Add extension slot to enable adding print button for a medication dispense

mseaton:
## Requirements

## Summary
Adds ability to add additional menu items to the medication dispense overflow menu

## Screenshots
image

## Related Issue

https://issues.openmrs.org/browse/O3-5690

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -21,4 +21,4 @@ export const PRESCRIPTIONS_TABLE_ENDPOINT = 'Encounter?_query=encountersWithMedi
 
 export const PRESCRIPTION_DETAILS_ENDPOINT = 'MedicationRequest';
 
-export const MEDICATION_DISPENSE_ACTION_MENU_ITEM_SLOT = 'medication-dispense-action-menu-item-slot';
+export const MEDICATION_DISPENSE_ACTION_SLOT = 'medication-dispense-action-slot';
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -20,3 +20,5 @@ export const JSON_MERGE_PATH_MIME_TYPE = 'application/merge-patch+json';
 export const PRESCRIPTIONS_TABLE_ENDPOINT = 'Encounter?_query=encountersWithMedicationRequests';
 
 export const PRESCRIPTION_DETAILS_ENDPOINT = 'MedicationRequest';
+
+export const MEDICATION_DISPENSE_ACTION_MENU_ITEM_SLOT = 'medication-dispense-action-menu-item-slot';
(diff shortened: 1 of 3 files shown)


denniskigen (commented):
LGTM.

Optional nits:
- `onOpen={() => setMenuOpen(true)}` + `onClose={() => setMenuOpen(false)}` is slightly more semantic than `onClick={toggle}` + `onClose`, since Carbon fires `onClick` only on the trigger button anyway. Functionally equivalent; just easier to read.

mseaton:
> LGTM

Thanks @denniskigen - is this approved for merging then?  Honestly I don't really love this approach, and I would like to see us consider whether this overflow menu is the right design in general as it adds another click where I would just want to have a "Print label" action that doesn't first require opening a menu.  I also don't really like how "Delete" - which is destructive and rarely what is wanted - is so visually and clickably close to print label - which is something one might want to do on every dispense.  So I would be very open to discussing a redesign of this.  For now though, I think just getting the ability to add extensible actions in place for dispenses is useful to unblock getting this tested.  Interested in others thoughts - @ibacher / @mogoodrich / @chibongho

mogoodrich:
Haven't looked at the code, but would be good there was an option/means to insert the print (or any option) at the top of the list, as this seems like a function that would be more frequently used that edit or delete.

mseaton:
> Haven't looked at the code, but would be good there was an option/means to insert the print (or any option) at the top of the list, as this seems like a function that would be more frequently used that edit or delete.

Yeah, I agree, but also am not sure these things all belong in the same list, and wanted to get that discussion going.  If we do keep them in the same list, the right thing is just making all of the actions into extensions, rather than hard-coding the edit and delete ones, and then the could be re-ordered in config.

I'm not trying to add comments here to hold up merging this work, more to indicate I think this approach is somewhat flawed and wanting to check how we might want to address that.

chibongho:
I still think having a slot next to the OverflowMenu to add a print button, or other more frequently used buttons, is cleaner and easier to click than putting the option within the menu, but I'll defer to BAs for the UX.

mseaton:
> I still think having a slot next to the OverflowMenu to add a print button, or other more frequently used buttons, is cleaner and easier to click than putting the option within the menu, but I'll defer to BAs for the UX.

I agree.  I have changed it around.  Here is what it currently looks like with my Print Label extension added from my custom esm:

image

How do we feel about this?  @denniskigen / @ibacher ?  And is this the right name/identifier for the slot?  Should we choose something more specific to this particular use case?

mogoodrich:
This works for me.  Not really a UX designer, but seems like separating the edit and delete functionality (not happy path/infrequent) and making the happy path options (print) more prominent does make sense to me.

chibongho (approved):
LGTM

PR #835, opened 2026-05-19
(fix) Restore webservices.rest 3.4.x compatibility

denniskigen:
## Requirements

## Summary

#832 raised the `webservices.rest` backend dependency from `>=2.2.0` to `>=3.5.0 <4.0.0` while adding the location-filter feature. Looking at that PR though, the only new backend requirement it actually introduced is the optional `emrapi` module, which is already gated behind the `emrapi-module` feature flag. Nothing in the dispensing app depends on a 3.5-only API, so the tighter floor needlessly blocks installs on backends that are still on 3.4.x.

This PR lowers the floor back to `>=3.4.0 <4.0.0` to restore compatibility.

## Screenshots

Missing backend modules validation error:

CleanShot 2026-05-19 at 11 34 00@2x

## Related Issue

N/A

## Other

N/A

### DIFF
--- a/src/routes.json
+++ b/src/routes.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.openmrs.org/routes.schema.json",
   "backendDependencies": {
     "fhir2": ">=1.2",
-    "webservices.rest": ">=3.5.0 <4.0.0"
+    "webservices.rest": ">=3.4.0 <4.0.0"
   },
   "optionalBackendDependencies": {
     "emrapi": {
--- a/src/routes.json
+++ b/src/routes.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.openmrs.org/routes.schema.json",
   "backendDependencies": {
     "fhir2": ">=1.2",
-    "webservices.rest": ">=3.4.0 <4.0.0"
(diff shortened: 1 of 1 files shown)


NethmiRodrigo:
Could it be to make sure that this - https://github.com/openmrs/openmrs-module-webservices.rest/pull/741 was included in the distro?

src/routes.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://json.openmrs.org/routes.schema.json",
   "backendDependencies": {
     "fhir2": ">=1.2",
-    "webservices.rest": ">=3.5.0 <4.0.0"
+    "webservices.rest": ">=3.4.0 <4.0.0"
```
chibongho:
```suggestion
    "webservices.rest": ">=2.2.0 <4.0.0"
```

chibongho:
Thanks for fixing this. I originally wanted to use the newly added `descendantLocations` properly in LocationResource in the REST API, but switched to using EMR API instead. We can revert back to the old REST version lower bound  (>=2.2.0) while keeping the newly added upper bound (<4.0.0).

chibongho (approved):

PR #836, opened 2026-05-20
(feat) O3-5650: Add ability to limit prescriptions to user's visit lo…

chibongho:
…cation

## Requirements

## Summary

Previous commit e03985d missed a requirement that when `restrictToVisitLocationDescendants` is configured to true, we should not only restrict the location filter options to descendants of the user session's visit location, but also further restrict prescription rows to only those locations, even if the user did not select anything in the location filter.  This PR fixes that.

Also lifted the `restrictToVisitLocationDescendants` config from `locationBehavior.locationFilter` to `locationBehavior` as it affects more than just the filter.

## Related Issue

https://issues.openmrs.org/browse/O3-5650


mogoodrich (approved):
LGTM, thanks @chibongho !

PR #843, opened 2026-06-03
(fix) O3-5650: Subject results in the search tab to location filterin…

chibongho:
…g config

## Requirements

## Summary

12a722e added a new config: `locationBehavior.restrictToVisitLocationDescendants`. When set to true, it should limit the dispensing rows we see to only those from locations under the session's visit location. This restriction was implemented for the "Active Prescription" and "All Prescription" tabs, but not the "Search" tab. This PR fixes that.

## Screenshots

The search tab now only shows prescriptions from sub-locations of the session's visit location.

image

### DIFF
--- a/src/prescriptions/prescription-tab-lists.component.tsx
+++ b/src/prescriptions/prescription-tab-lists.component.tsx
@@ -66,7 +66,7 @@ const PrescriptionTabLists: React.FC = () => {
             ))}
           </TabList>
           <TabPanels>
-            <PatientSearchTabPanel />
+            <PatientSearchTabPanel locations={locations} />
             <PrescriptionTabPanel
               isTabActive={selectedTab === 1}
               status={'ACTIVE'}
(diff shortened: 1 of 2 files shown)


mseaton (approved):

src/prescriptions/prescription-tab-lists.component.tsx
```diff
@@ -66,7 +66,7 @@ const PrescriptionTabLists: React.FC = () => {
             ))}
           </TabList>
           <TabPanels>
-            <PatientSearchTabPanel />
+            <PatientSearchTabPanel locations={locations} />
```
denniskigen, line 69: Thanks for wiring the search tab into the existing location plumbing, @chibongho.

Could we gate the locations passed here on `config.locationBehavior.restrictToVisitLocationDescendants`? `useLocations(config)` returns the tag-filtered location list when that setting is false, and `PrescriptionsTable` turns any non-empty list into `&location=...`, which FHIR2 applies as an encounter location filter. With the default config, Search would now be restricted to tagged locations instead of remaining unrestricted as before. For example, a prescription ordered at a location without the `Login Location` tag no longer shows up in search results.

Something like:

```tsx
<PatientSearchTabPanel
  locations={config.locationBehavior.restrictToVisitLocationDescendants ? locations : []}
/>
```

This would also keep Search consistent with the other tab panels, which already gate on this same flag.

PR #854, opened 2026-06-11
(fix) O3-5734: Allow recording allergies from the fill prescription order basket

denniskigen:
## Requirements

## Summary

The "Add allergy" (+) control in the order basket's drug search falls back to launching the patient chart's allergy form workspace when the host doesn't provide a launcher. The fill prescription workflow renders the order basket outside the patient chart, so that fallback opened the `patient-chart` workspace group with null group props and crashed every action button registered to it with `Cannot read properties of null (reading 'patientUuid')`.

This registers the exported allergy form workspace (added in openmrs/openmrs-esm-patient-chart#3353) into the `fill-prescription-window` and passes its name as `allergyFormWorkspaceName` in the order basket window props, so the form opens as a child of the dispensing app's own workspace window. This is the same wiring the ward app added in openmrs/openmrs-esm-patient-management#2561.

Note: this requires the patient chart release that ships `exportedAllergyFormWorkspace` (v12.2.0 or higher).

## Screenshots

https://github.com/user-attachments/assets/c25747d4-3f81-484a-87e2-7c4c9c1c396a

## Related Issue

https://openmrs.atlassian.net/browse/O3-5734

### DIFF
--- a/src/fill-prescription/fill-prescription-button.component.tsx
+++ b/src/fill-prescription/fill-prescription-button.component.tsx
@@ -47,6 +47,7 @@ const FillPrescriptionButton: React.FC<{}> = () => {
                   patient: patient,
                   visitContext: activeVisit,
                   drugOrderWorkspaceName: 'dispensing-order-basket-add-drug-order-workspace',
+                  allergyFormWorkspaceName: 'dispensing-order-basket-add-allergy-workspace',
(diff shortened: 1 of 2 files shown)


NethmiRodrigo (approved):
Thanks @denniskigen!

PR #859, opened 2026-06-22
(fix) Refresh prescriptions table in place after dispensing

NethmiRodrigo:
## Requirements

## Summary

When dispensing (or pausing/closing/deleting) a prescription, the prescriptions table did not update in place — the status and last-dispenser only changed after switching tabs or manually refreshing.

The root cause is a SWR cache mismatch. `revalidate()` in `src/utils.ts` used the **global `mutate` imported from `swr`**, which is permanently bound to SWR's *default* cache. OpenMRS wraps every microfrontend in an `<SWRConfig>` with a **custom cache provider** (`@openmrs/esm-react-utils` → `provider: () => swrCache`), so all `useSWR` data — including the prescriptions table — lives in that provider's cache. The global `mutate`'s key filter therefore matched nothing and the table entry was never revalidated.

Switching tabs appeared to "fix" it only because `loadData={isTabActive}` flips the SWR key (`null` ↔ URL), which forces a fresh fetch independently of `mutate`.

### Fix

Pass the cache-bound `mutate` from `useSWRConfig()` into `revalidate()` so it targets the active provider's cache:

- `revalidate(encounterUuid)` → `revalidate(mutate, encounterUuid)`
- Updated all callers: dispense / close / pause forms, the on-prescription-filled modal, and the delete handler in `history-and-comments`.

## Screenshots

Before - 

https://github.com/user-attachments/assets/9b5c9a2a-9948-49ca-9f21-e03d5dffb85f

After - 

https://github.com/user-attachments/assets/51990da6-9a9b-4968-a27b-6a9e8684b6a1

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,6 +1,6 @@
 import dayjs from 'dayjs';
 import template from 'lodash/template';
-import { mutate } from 'swr';
+import { type ScopedMutator } from 'swr/_internal';
 import {
   type Coding,
   type DispensingStore,
@@ -580,10 +580,16 @@ export function isMostRecentMedicationDispense(
 
 /**
  * Revalidated (reloads) both the prescription associated with the encounter uuid,
- * and the entire prescription table
+ * and the entire prescription table.
+ *
+ * The `mutate` must be the one obtained from `useSWRConfig()` in the calling component, not the
+ * global `mutate` imported from `swr`. OpenMRS wraps every microfrontend in an `<SWRConfig>` with a
+ * custom cache provider, so the global `mutate` (bound to SWR's default cache) does not see the
+ * app's cached entries and would silently no-op.
+ * @param mutate the cache-bound mutator from `useSWRConfig()`
  * @param encounterUuid
  */
-export async function revalidate(encounterUuid: string) {
+export async function revalidate(mutate: ScopedMutator, encounterUuid: string) {
   await mutate(
     (key) =>
       typeof key === 'string' &&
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -1,6 +1,6 @@
 import dayjs from 'dayjs';
 import template from 'lodash/template';
-import { mutate } from 'swr';
+import { type ScopedMutator } from 'swr/_internal';
```
ibacher: `ScopedMutator` should be re-exported from `'swr'` and I think it's preferable to use that version.

ibacher (approved):
One small thing, but LGTM otherwise.

src/utils.ts
denniskigen: On the pinned swr 2.2.5, `ScopedMutator` isn't part of the public `swr` re-export (only `swr/_internal` exports it), so `import { type ScopedMutator } from 'swr'` fails to type-check with `TS2614`. I confirmed by compiling both imports against the repo's tsconfig, so the current `swr/_internal` import is the only one that resolves in this version.

If the goal is to avoid the `_internal` deep-import, deriving the type from the public API works without it:

```ts
type ScopedMutator = ReturnType<typeof useSWRConfig>['mutate'];
```

denniskigen (approved):

src/utils.ts
ibacher: Fair. Then it's better to use the syntax suggested here. We really shouldn't be directly invoking `swr/_internal`.

NethmiRodrigo: Done

PR #889, opened 2026-08-20
(feat) O3-5907: Show medication side effects on prescription details

druchniewicz:
## Requirements

 - [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#contributing-guidelines) label. See existing PR titles for inspiration.

## Summary

Mounts a `dispensing-prescription-side-effects-slot` on the prescription details tile, guarded on the drug uuid parsed from the request's `medicationReference`, so the `openmrs-esm-medication-side-effects` microfrontend can render its panel per prescribed drug. 

## Screenshots

image

## Related Issue

  https://openmrs.atlassian.net/browse/O3-5907

## Other

  Related PRs:
  * https://github.com/openmrs/openmrs-esm-medication-side-effects/pull/1
  * https://github.com/openmrs/openmrs-esm-patient-chart/pull/3591       
  * https://github.com/openmrs/openmrs-module-medicationsideeffects/pull/1
  * https://github.com/mekomsolutions/openmrs-module-initializer/pull/333
  * https://github.com/openmrs/openmrs-content-referenceapplication-demo/pull/85

### DIFF
--- a/src/components/medication-event.component.tsx
+++ b/src/components/medication-event.component.tsx
@@ -21,8 +21,10 @@ const MedicationEvent: React.FC<{
   medicationEvent: MedicationRequest | MedicationDispense;
   status?: ReactNode;
   children?: ReactNode;
+  footer?: ReactNode;
+  alignContentStart?: boolean;
   isDispenseEvent?: boolean;
-}> = ({ medicationEvent, status = null, children, isDispenseEvent }) => {
+}> = ({ medicationEvent, status = null, children, footer = null, alignContentStart = false, isDispenseEvent }) => {
   const { t } = useTranslation();
   const dosageInstruction = getDosageInstruction(medicationEvent.dosageInstruction);
   const isFreeTextDosage = calculateIsFreeTextDosage(dosageInstruction);
@@ -37,58 +39,61 @@ const MedicationEvent: React.FC<{
         [styles.dispenseEvent]: isDispenseEvent,
         [styles.isTablet]: isTablet,
       })}>
-      <div>
-        <p className={styles.medicationName}>
-          {status}
-          {status && ' '}
-          <strong>{getMedicationDisplay(getMedicationReferenceOrCodeableConcept(medicationEvent))}</strong>
-        </p>
-
-        {!isFreeTextDosage && (
-          <p className={styles.bodyLong01}>
-            <span className={styles.label01}>{t('dose', 'Dose').toUpperCase()}</span>{' '}
-            <span className={styles.dosage}>
-              {dosageInstruction?.doseAndRate &&
-                dosageInstruction.doseAndRate.map((doseAndRate, index) => {
-                  return (
-                    <span key={index}>
-                      {doseAndRate?.doseQuantity?.value} {doseAndRate?.doseQuantity?.unit}
-                    </span>
-                  );
-                })}
-            </span>
-            {dosageInstruction?.route?.text && <> &mdash; {dosageInstruction.route.text}</>}
-            {dosageInstruction?.timing?.code?.text && <> &mdash; {dosageInstruction.timing.code.text}</>}
-            {dosageInstruction?.timing?.repeat?.duration && (
-              <>
-                {' '}
-                for {dosageInstruction.timing.repeat.duration} {dosageInstruction.timing.repeat.durationUnit}
-              </>
-            )}
+      <div className={classNames(styles.content, { [styles.contentStart]: alignContentStart })}>
+        <div>
+          <p className={styles.medicationName}>
+            {status}
+            {status && ' '}
+            <strong>{getMedicationDisplay(getMedicationReferenceOrCodeableConcept(medicationEvent))}</strong>
           </p>
-        )}
 
-        {quantity && (
-          <p className={styles.bodyLong01}>
-            <span className={styles.label01}>{t('quantity', 'Quantity').toUpperCase()}</span>{' '}
-            <span className={styles.quantity}>
-              {quantity.value} {quantity.unit}
-            </span>
-          </p>
-        )}
+          {!isFreeTextDosage && (
+            <p className={styles.bodyLong01}>
+              <span className={styles.label01}>{t('dose', 'Dose').toUpperCase()}</span>{' '}
+              <span className={styles.dosage}>
+                {dosageInstruction?.doseAndRate &&
+                  dosageInstruction.doseAndRate.map((doseAndRate, index) => {
+                    return (
+                      <span key={index}>
+                        {doseAndRate?.doseQuantity?.value} {doseAndRate?.doseQuantity?.unit}
+                      </span>
+                    );
+                  })}
+              </span>
+              {dosageInstruction?.route?.text && <> &mdash; {dosageInstruction.route.text}</>}
+              {dosageInstruction?.timing?.code?.text && <> &mdash; {dosageInstruction.timing.code.text}</>}
+              {dosageInstruction?.timing?.repeat?.duration && (
+                <>
+                  {' '}
+                  for {dosageInstruction.timing.repeat.duration} {dosageInstruction.timing.repeat.durationUnit}
+                </>
+              )}
+            </p>
+          )}
 
-        {(refillsAllowed || refillsAllowed === 0) && (
-          <p className={styles.bodyLong01}>
-            <span className={styles.label01}>{t('refills', 'Refills').toUpperCase()}</span>{' '}
-            <span className={styles.refills}>{refillsAllowed}</span>
-          </p>
-        )}
-        {dosageInstruction?.text && <p className={styles.bodyLong01}>{dosageInstruction.text}</p>}
-        {dosageInstruction?.additionalInstruction?.length > 0 && (
-          <p className={styles.bodyLong01}>{dosageInstruction?.additionalInstruction[0].text}</p>
-        )}
+          {quantity && (
+            <p className={styles.bodyLong01}>
+              <span className={styles.label01}>{t('quantity', 'Quantity').toUpperCase()}</span>{' '}
+              <span className={styles.quantity}>
+                {quantity.value} {quantity.unit}
+              </span>
+            </p>
+          )}
+
+          {(refillsAllowed || refillsAllowed === 0) && (
+            <p className={styles.bodyLong01}>
+              <span className={styles.label01}>{t('refills', 'Refills').toUpperCase()}</span>{' '}
+              <span className={styles.refills}>{refillsAllowed}</span>
+            </p>
+          )}
+          {dosageInstruction?.text && <p className={styles.bodyLong01}>{dosageInstruction.text}</p>}
+          {dosageInstruction?.additionalInstruction?.length > 0 && (
+            <p className={styles.bodyLong01}>{dosageInstruction?.additionalInstruction[0].text}</p>
+          )}
+        </div>
+        {children}
       </div>
-      {children}
+      {footer && <div className={styles.footer}>{footer}</div>}
     </Tile>
   );
 };
--- a/src/prescriptions/prescription-details.component.tsx
+++ b/src/prescriptions/prescription-details.component.tsx
@@ -2,7 +2,13 @@ import React from 'react';
 import { SkeletonText, Tag, Tile } from '@carbon/react';
 import { WarningFilled } from '@carbon/react/icons';
 import { useTranslation } from 'react-i18next';
-import { ExtensionSlot, getAssignedExtensions, type PatientUuid, useConfig, UserHasAccess } from '@openmrs/esm-framework';
+import {
+  ExtensionSlot,
+  getAssignedExtensions,
+  type PatientUuid,
+  useConfig,
+  UserHasAccess,
+} from '@openmrs/esm-framework';
 import {
   computeMedicationRequestCombinedStatus,
   getConceptCodingDisplay,
@@ -163,9 +169,7 @@ const PrescriptionDetails: React.FC<{
                 status={generateStatusTag(bundle)}
                 alignContentStart
                 footer={actionButtons}>
-                {drugUuid && (
-                  <ExtensionSlot name="dispensing-prescription-side-effects-slot" state={{ drugUuid }} />
-                )}
+                {drugUuid && <ExtensionSlot name="dispensing-prescription-side-effects-slot" state={{ drugUuid }} />}
               </MedicationEvent>
             ) : (
               <MedicationEvent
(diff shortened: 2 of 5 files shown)


druchniewicz:
@denniskigen @ibacher @dkayiwa can you take a look?

druchniewicz:
@denniskigen @ibacher @dkayiwa bumping this

dkayiwa (commented):
I don't think anything in the diff blocks this, though I'd like an answer on the layout question before it goes in.

What is actually holding the merge is the two red required checks rather than the code. The title needs a type prefix for the frontend PR title check, so `(feat) O3-5907: Show medication side effects on prescription details`, and the description needs to use the repo template, with the Requirements, Summary, Screenshots, Related Issue and Other headings.

src/prescriptions/prescription-details.component.tsx
```diff
@@ -145,15 +146,22 @@ const PrescriptionDetails: React.FC<{
...
-              <UserHasAccess privilege={PRIVILEGE_CREATE_DISPENSE}>
-                <ActionButtons
-                  patientUuid={patientUuid}
-                  encounterUuid={encounterUuid}
-                  medicationRequestBundle={bundle}
-                  disabled={staleEncounterUuids.includes(encounterUuid)}
-                />
-              </UserHasAccess>
+              status={generateStatusTag(bundle)}
+              alignContentStart
```
dkayiwa: Was the new tile layout meant to apply to implementations that haven't installed the medication side effects app? Right now it does, and it costs them something. I compiled the base and PR stylesheets and measured the tile in headless Chromium with the slot empty: it grows by 64px (the action button row plus the footer's 1rem top margin), the Close/Pause/Dispense buttons move from beside the drug details down to a row of their own, and the right half of the tile is left blank.

If that isn't wanted, the parent component already has the pattern: `prescription-expanded.component.tsx` gates its Conditions and diagnoses tab on `getAssignedExtensions('dispensing-condition-and-diagnoses')`, and the same check here would leave the layout as it is today for anyone who hasn't installed the app. It wouldn't cover the case where the app is installed but a drug has no side effect rows, since the panel returns null then and the tile is still 64px taller, so keeping one layout everywhere is a fair answer too. I just wouldn't want it decided by accident.

src/prescriptions/prescription-details.component.tsx
```diff
@@ -145,15 +146,22 @@ const PrescriptionDetails: React.FC<{
...
+                    patientUuid={patientUuid}
+                    encounterUuid={encounterUuid}
+                    medicationRequestBundle={bundle}
+                    disabled={staleEncounterUuids.includes(encounterUuid)}
+                  />
+                </UserHasAccess>
+              }>
+              <ExtensionSlot
+                name="dispensing-prescription-side-effects-slot"
+                state={{ drugUuid: getUuidFromReference(bundle.request.medicationReference?.reference) }}
```
dkayiwa: Worth guarding this on the uuid actually being there. A request that carries only a `medicationCodeableConcept`, which `getMedicationDisplay` in `utils.ts` still handles, has no `medicationReference`, so the slot gets mounted with `drugUuid: undefined`. I rendered `PrescriptionDetails` with such a request to check, and it does mount, with an empty state object.

Nothing shows today because the panel in openmrs-esm-medication-side-effects returns null on a falsy `drugUuid`, but that is that extension's own behaviour rather than part of the slot contract, and any other extension someone plugs in here inherits the guard. The patient chart half of this ticket does gate it (`{drug?.uuid && <ExtensionSlot name="order-form-side-effects-slot" ... />}`), so guarding here would keep the two hosts consistent.

src/components/medication-event.component.tsx
```diff
@@ -37,58 +39,61 @@ const MedicationEvent: React.FC<{
...
+          )}
+          {dosageInstruction?.text && <p className={styles.bodyLong01}>{dosageInstruction.text}</p>}
+          {dosageInstruction?.additionalInstruction?.length > 0 && (
+            <p className={styles.bodyLong01}>{dosageInstruction?.additionalInstruction[0].text}</p>
+          )}
+        </div>
+        {children}
       </div>
-      {children}
+      {footer && <div className={styles.footer}>{footer}</div>}
```
dkayiwa, line 96: This tests whether `footer` was passed rather than whether it renders anything, and at the one call site it's always passed: `<UserHasAccess>` is a truthy element even when it renders null. So the wrapper and its `margin-top: 1rem` are emitted on every row for a user without `Task: dispensing.create.dispense`, and for any prescription where none of Dispense, Pause or Close applies, since all three extensions return null once the request is no longer active. Measured, that's 16px of dead space at the bottom of the tile in both cases.

`.footer:empty { display: none; }` would cover the privilege case. The non-active case needs the wrapper itself not to render, because `ActionButtons` still emits `.actionBtns` and the slot div even when no button does.

src/components/medication-event.component.tsx
```diff
@@ -21,8 +21,10 @@ const MedicationEvent: React.FC<{
   medicationEvent: MedicationRequest | MedicationDispense;
   status?: ReactNode;
   children?: ReactNode;
+  footer?: ReactNode;
+  alignContentStart?: boolean;
```
dkayiwa, line 25: These two props, and the slot they were added for, aren't covered by a test, though both components already have suites: `medication-event.test.tsx` has a "renders children when provided" case and nothing equivalent for `footer`, and `prescription-details.test.tsx` only exercises the allergies panel. The repo's PR template also asks for tests or existing coverage. Since the slot name and the `drugUuid` state shape are now a contract openmrs-esm-medication-side-effects depends on, pinning them is worth it so that a rename here shows up as a red test rather than as an empty panel over there.

I wrote a throwaway spec on this branch asserting that `ExtensionSlot` is called with `dispensing-prescription-side-effects-slot` and `{ drugUuid: 'drug-uuid-1' }` for a `Medication/drug-uuid-1` reference, plus one that the footer wrapper renders. Both pass here and fail on main, so the existing fixtures are enough to write them. `action-buttons.test.tsx` already has the `mockedExtensionSlot` pattern to copy.

src/prescriptions/prescription-details.component.tsx
druchniewicz: Fixed. The footer/top-aligned layout is now used only when an extension is assigned to dispensing prescription-side-effects-slot (via getAssignedExtensions), the same check prescription-expanded uses. Implementations without the medication side effects app keep the original tile layout.

src/prescriptions/prescription-details.component.tsx
druchniewicz: fixed, the slow now mounts only when the uuid parsed from medicationReference is present

src/components/medication-event.component.tsx
druchniewicz: Added .footer:empty { display: none; }

src/components/medication-event.component.tsx
druchniewicz: tests added

druchniewicz:
@dkayiwa changes applied

src/prescriptions/prescription-details.test.tsx
```diff
@@ -1,12 +1,19 @@
...
-import { useConfig } from '@openmrs/esm-framework';
+import { ExtensionSlot, getAssignedExtensions, useConfig } from '@openmrs/esm-framework';
 import { usePrescriptionDetails, usePatientAllergies } from '../medication-request/medication-request.resource';
 import { useStaleEncounterUuids } from '../utils';
 import type * as Utils from '../utils';
+import { type MedicationRequest, type MedicationRequestBundle, MedicationRequestStatus } from '../types';
 import PrescriptionDetails from './prescription-details.component';
 
+vi.mock('@openmrs/esm-framework', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('@openmrs/esm-framework')>()),
```
dkayiwa: This one has to be fixed before the PR can go in: `eslint src` fails on this commit, so the Build job goes red as soon as someone approves the workflow run (it's sitting at `action_required` right now, which is why the PR page doesn't show it yet). The shared build workflow runs `yarn verify --concurrency=5`, which is `turbo run lint test typescript`, so lint is a hard gate.

```
 12:35  error  `import()` type annotations are forbidden       @typescript-eslint/consistent-type-imports
337:31  error  This assertion is unnecessary since the
               receiver accepts the original type              @typescript-eslint/no-unnecessary-type-assertion
```

Both arrived with this commit; 5f5a16c lints clean. The first wants the shape the file already uses for `../utils` a couple of lines up:

```suggestion
import type * as EsmFramework from '@openmrs/esm-framework';
import type * as Utils from '../utils';
import { type MedicationRequest, type MedicationRequestBundle, MedicationRequestStatus } from '../types';
import PrescriptionDetails from './prescription-details.component';

vi.mock('@openmrs/esm-framework', async (importOriginal) => ({
  ...(await importOriginal<typeof EsmFramework>()),
```

For the second, the `as Partial<MedicationRequest>` in the "without a medication reference" test can just go, since `buildRequest` already declares that parameter type. With both applied, `eslint src`, `tsc --noEmit` and this file's 12 tests all pass here.

Prettier also rewrites `prescription-details.component.tsx` (the `@openmrs/esm-framework` import is 122 characters against a printWidth of 120, and the `ExtensionSlot` line fits on one). Not a CI gate, just worth folding in while you're in there.

dkayiwa (commented):
Not mergeable yet, though nothing in the behaviour of the change is the reason.

The description check is still red, and it's only whitespace. All five headings carry two leading spaces (`  ## Requirements`, `  ## Summary`, and so on), while the validator matches `^#{1,6}\s+<name>\s*$` line by line, so an indented heading doesn't count and it reports all five as missing. I ran that check against the current body: unindent those five lines and it passes, since the first-checkbox and Summary-content rules are already satisfied.

That, plus the lint failure in the comment on the test file, is what's left before merge.

src/components/medication-event.component.tsx
dkayiwa: The `:empty` rule covers the privilege case, but the other half of this is still there. `ActionButtons` always returns its `.actionBtns` wrapper around the extension slot div, with no early return, so when the user does have `Task: dispensing.create.dispense` but the request is no longer active and all three buttons render null, `.footer` has children and `:empty` doesn't match. Measured against the compiled stylesheet in headless Chromium, the footer box collapses to 0 height but its `margin-top: 1rem` still applies, so the tile comes out 184.7px against 168.7px with the footer gone.

That's 16px of trailing whitespace on non-actionable prescriptions, and only where the side effects app is installed, so I'd leave it rather than have `ActionButtons` return null when none of dispensable/pauseable/closeable holds, which would also hide any button an implementation has attached to `prescription-action-button-slot`. Worth deciding rather than assuming `:empty` caught both cases.

src/prescriptions/prescription-details.test.tsx
druchniewicz: fixed

src/components/medication-event.component.tsx
druchniewicz: Leaving it, since removing it would mean ActionButtons early-returning null, which would also hide button implementations attach to prescription-action-button-slot

druchniewicz:
@dkayiwa changes applied

src/prescriptions/prescription-details.component.tsx
```diff
@@ -37,6 +44,7 @@ const PrescriptionDetails: React.FC<{
   } = usePatientAllergies(patientUuid, config.refreshInterval);
   const { medicationRequestBundles, error, isLoading } = usePrescriptionDetails(encounterUuid, config.refreshInterval);
   const { staleEncounterUuids } = useStaleEncounterUuids();
+  const sideEffectsSlotFilled = getAssignedExtensions('dispensing-prescription-side-effects-slot').length > 0;
```
dkayiwa: This line throws on any install that doesn't have `openmrs-esm-medication-side-effects`, which right now is every install, since that app's own PR is still open. It needs fixing before this can merge.

`getAssignedExtensions` reads `internalState.slots[slotName].attachedIds` with no guard (esm-extensions, `getAssignedExtensionsFromSlotData`), and a slot only enters that store when something attaches to it: an extension naming it in its routes.json, or an `<ExtensionSlot>` of that name mounting. Nothing in this repo attaches to `dispensing-prescription-side-effects-slot`, and the slot below is gated behind this very call, so on those installs the entry is missing and this raises `TypeError: Cannot read properties of undefined (reading 'attachedIds')` during render. `openmrsComponentDecorator`'s boundary catches it and swaps the whole dispensing page for "An error has occurred. Please try reloading the page.", so expanding any prescription row is enough to lose the page.

I registered this repo's own routes.json the way `@openmrs/esm-routes` does (`registerExtension`, then `attach` per slot) and rendered `PrescriptionDetails` against the real extension store: it throws here. In the same run `getAssignedExtensions('dispensing-condition-and-diagnoses')` comes back with `patient-diagnoses` and `patient-conditions`, which is why the same call in `prescription-expanded.component.tsx` is safe. Those two are attached by this repo's own routes.json, so that slot always exists. The suite stays green because `prescription-details.test.tsx` replaces `getAssignedExtensions` with `vi.fn(() => [])`, so the lookup that throws never runs.

`useAssignedExtensions` gives you the same list. It reads the output store that `getAssignedExtensionsFromSlotData` fills, so privileges, feature flag, display expression and online/offline still apply, and the store is recomputed when the session loads. The difference is that it is `slotStore?.assignedExtensions ?? []`, so a missing slot yields `[]` instead of throwing, and being a subscription it also picks up an extension that registers after the first render.

```suggestion
  const sideEffectsSlotFilled = useAssignedExtensions('dispensing-prescription-side-effects-slot').length > 0;
```

with the import above changed to match. In the spec that is `mockUseAssignedExtensions` in place of `mockGetAssignedExtensions`, and the `vi.mock('@openmrs/esm-framework', ...)` factory can go altogether, since the framework's own mock already exports `useAssignedExtensions` as `vi.fn(() => [])`. With that applied, `eslint src`, `tsc` and all 167 tests pass here, and the render above stops throwing.

src/prescriptions/prescription-details.component.tsx
dkayiwa: The tile itself matches main now, with one line left over. `.content` in medication-event.scss carries `gap: layout.$spacing-05`, which main's `.medicationEventTile` never had (it was `space-between` with no gap), and it applies on the default path too. Measuring the compiled stylesheets in headless Chromium, that costs the drug details column 16px whenever the row is full: at an 800px tile the details box goes from 524.7px to 508.7px, heights unchanged. Dropping that one declaration restores base geometry exactly at every width I tried, and changes nothing where the app is installed, since `.contentStart` sets its own `column-gap` and tablet still takes its gap from `.isTablet .content`. Not a blocker, just the last part of the restructure that isn't neutral for implementations without the app.

druchniewicz:
@dkayiwa changes applied

PR #895, opened 2026-09-01
(feat) O3-5942: Sort renewed prescriptions using the latest MedicationRequest authored date

donaldkibet:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#contributing-guidelines) label. See existing PR titles for inspiration.

## Summary
**Problem**
The Search, Active Prescriptions, and All Prescriptions tables used Encounter.period.start as the prescription created date and sort value.

When a medication is renewed within an existing encounter, OpenMRS creates a new MedicationRequest with a new authoredOn date. Because the tables still referenced the original encounter date, renewed prescriptions displayed an outdated date and could be sorted below older prescriptions.

**Change**
Updated the prescription tables to use MedicationRequest.authoredOn as the created date and sorting value.

## Screenshots
https://github.com/user-attachments/assets/d9f99149-1490-4a15-8769-1e9abf6d14ce

### DIFF
--- a/src/medication-request/medication-request.resource.tsx
+++ b/src/medication-request/medication-request.resource.tsx
@@ -106,7 +106,7 @@ function buildPrescriptionsTableRow(
 ): PrescriptionsTableRow {
   return {
     id: encounter?.id,
-    created: encounter?.period?.start,
+    created: getPrescriptionRowDate(encounter, medicationRequests),
     patient: {
       name: encounter?.subject?.display,
       uuid: encounter?.subject?.reference?.split('/')[1],
@@ -126,6 +126,16 @@ function buildPrescriptionsTableRow(
     status: computePrescriptionStatusMessageCode(medicationRequests, medicationRequestExpirationPeriodInDays),
     location: encounter?.location ? encounter?.location[0]?.location.display : null,
   };
+}
+
+export function getPrescriptionRowDate(encounter: Encounter, medicationRequests: Array<MedicationRequest>): string {
+  const newestAuthoredOn = medicationRequests
+    .map((medicationRequest) => medicationRequest.authoredOn)
+    .filter((authoredOn): authoredOn is string => Boolean(authoredOn) && dayjs(authoredOn).isValid())
+    .sort((a, b) => dayjs(b).valueOf() - dayjs(a).valueOf())[0];
+
+  // Older MedicationRequests may not have authoredOn, so retain the encounter date as a fallback.
+  return newestAuthoredOn ?? encounter?.period?.start;
 }
 
 export function usePrescriptionDetails(encounterUuid: string, refreshInterval = null) {
--- a/src/medication-request/medication-request.resource.test.tsx
+++ b/src/medication-request/medication-request.resource.test.tsx
@@ -4,6 +4,7 @@ import dayjs from 'dayjs';
 import useSWR from 'swr';
 import {
   updateMedicationRequestFulfillerStatus,
+  getPrescriptionRowDate,
   useMedicationRequest,
   usePatientAllergies,
   usePrescriptionDetails,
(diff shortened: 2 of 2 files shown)


enyaencha (approved):

patryllus (approved):
LGTM thanks @donaldkibet

makombe (approved):
Thanks @donaldkibet

src/medication-request/medication-request.resource.tsx
```diff
@@ -106,7 +106,7 @@ function buildPrescriptionsTableRow(
 ): PrescriptionsTableRow {
   return {
     id: encounter?.id,
-    created: encounter?.period?.start,
+    created: getPrescriptionRowDate(encounter, medicationRequests),
```
denniskigen, line 109: I agree with the Greptile finding here. For the existing-encounter renewal this PR is handling, Search and Active still [send an encounter-date cutoff, while all three default tables send pagination parameters](https://github.com/openmrs/openmrs-esm-dispensing-app/blob/44beb140a0c3ba65f229a7e6566fbc116ab3bfc1/src/utils.ts#L496-L507). FHIR2 [applies the cutoff to the Encounter query](https://github.com/openmrs/openmrs-module-fhir2/blob/2be95ce6b90619a620a6a5ddba8dc4a7808ba0a9/api/src/main/java/org/openmrs/module/fhir2/providers/r4/EncounterFhirResourceProvider.java#L211-L214), [sorts Encounters by date](https://github.com/openmrs/openmrs-module-fhir2/blob/2be95ce6b90619a620a6a5ddba8dc4a7808ba0a9/api/src/main/java/org/openmrs/module/fhir2/providers/r4/EncounterFhirResourceProvider.java#L232-L236), and [paginates those sorted Encounter IDs](https://github.com/openmrs/openmrs-module-fhir2/blob/2be95ce6b90619a620a6a5ddba8dc4a7808ba0a9/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseFhirDao.java#L208-L220) before [adding the included MedicationRequests](https://github.com/openmrs/openmrs-module-fhir2/blob/2be95ce6b90619a620a6a5ddba8dc4a7808ba0a9/api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryBundleProvider.java#L77-L83). This change therefore only reorders one server page. An encounter older than the cutoff can still be absent from Search and Active, and a renewal whose encounter lands on a later All Prescriptions page cannot be promoted to page 1 by its newer `authoredOn`. Can we align the endpoint’s eligibility, ordering, and pagination with `authoredOn` before merging? If that is not possible server-side, the client would need the complete matching encounter set before applying the authored-date rule.

denniskigen (changes requested):
Good start, @donaldkibet. Using `authoredOn` fixes the date and ordering for rows that are already loaded. I’ve left one question about the remaining filtering and pagination.

mogoodrich (changes requested):
See my comments on https://openmrs.atlassian.net/browse/O3-5942 ... let's discuss there first.

My initial reaction is that (iirc) the Dispensing  App is tightly bound to the idea of all the orders associated with an encounter having actually been placed (or renewed) during that encounter. If we are changing that pattern, we may need to do a bigger review of how the Dispensing App works.

mogoodrich (changes requested):
So per @ibacher 's comment on O3-5942, we should not need to make this change, we should close this PR.

ibacher:
Per comments on the ticket, the fix here should be to ensure that renewals are recorded in an updated encounter which should then surface correctly in the dispensing app.

